Issue #38, opened 2026-04-17
[Discussion] Exploring ideas for expanding global CLI access, shims, and lifecycle management


bsirak:
**Description:**

First, thank you for the incredible work on `uvr`. I've been using it more and more on my projects and really love where it is right now. It’s a huge breath of fresh air compared to other tools on Windows (like `rig`) that require Administrator privileges, lack any real user control, and don't work together.

Because `uvr` already has a great foundation, I wanted to open a discussion about potentially expanding how those isolated `R` versions are exposed and managed. I have a few ideas on how `uvr` might evolve into THE comprehensive `R` tool that are heavily inspired by how Astral's `uv` handles Python, but since my specific workflow might not align with everyone's and because these concepts are interconnected, I wanted to put them all in one thread for feedback before opening individual feature requests.

**1. Expanding lifecycle commands (`update` and `uninstall`)**

Currently, `uvr` handles installation but lacks the ability to fully manage the lifecycle of those installations—most notably lacking a native way to uninstall or update managed versions.

- Proposing the addition of native `uvr r uninstall <version>` and `uvr r update` commands.

- Having the tool natively manage the full lifecycle (especially updating to the latest patch version) is a prerequisite for the auto-updating aliases proposed below.

**2. Global Access & Semantic Shims**

It would be amazing to invoke `uvr`-managed versions globally outside of an active project context.

- `uvr` could expose a dedicated user binary directory (e.g., `~/.uvr/bin/`) to the User `PATH`.

- Inside that folder, `uvr` could generate semantic aliases. For example, if `R 4.5.3` ( installed with a -`-default` flag or set via a `uvr r default <ver>` command) and `4.4.3` are installed, it automatically generates `R.exe` (pointing to `4.5.3`), `R4.5.exe`, and `R4.4.exe`.

- **Implementation**: To avoid `.bat` file overhead on Windows, `uvr` should use native, pre-compiled executable shims (`uv`'s trampolines) instead of batch scripts. On Unix, standard symlinks should work perfectly.

- **Auto-Updating**: When `uvr r update` is run, `uvr` automatically repoints the relevant high-level shims (like `R4.5.exe`) to the newest patch.

**3. An Interactive Console Wrapper (`uvr r run`)**

On Windows, a pre-existing System install (e.g., `C:\Program Files\R`) will preempt the user-space shims mentioned above if the user just types `R.exe`, because the OS evaluates the System `PATH` first.

[...]

PR #79, opened 2026-05-02
Feat: Add environment variables for path customization and centralize config

bsirak:
**Description:**
This PR introduces the ability to customize core `uvr` directory paths via environment variables (similar to how `uv` handles them) and centralizes all environment variable reads into a single ``config.rs`` module for better maintainability and documentation.

**New Features:**
Added support for overriding previously hardcoded paths using the following new environment variables:
- `UVR_CACHE_DIR`: Customizes the directory where packages, environments, and tarballs are cached (default: `~/.uvr/cache/`).
- `UVR_R_VERSIONS_DIR`: Customizes the directory where managed R versions are installed (default: `~/.uvr/r-versions/`).
- `UVR_INSTALL_DIR`: Customizes the directory where the `uvr` binary itself is installed during a `self-update`.

**Refactoring & Consolidation:**
To keep configuration consistent, previously existing environment variables scattered throughout the codebase have been moved into ``crates/uvr-core/src/config.rs``:
- `UVR_EXTRA_LIBS` *(moved from ``run.rs``)*
- `UVR_INSTALL_TIMEOUT` *(moved from ``r_cmd_install.rs``)*
- `UVR_LIBRARY` *(moved from ``sync.rs``)*
- `UVR_PROGRESS` *(moved from ``progress.rs``)*

**Changes:**
- Created ``crates/uvr-core/src/config.rs`` with alphabetically sorted accessor functions for all `UVR_*` variables.
- Added detailed docstrings for each variable, explaining expected formats and default behaviors.
- Updated all call sites across the codebase (e.g., ``manager.rs``, ``downloader.rs``, ``cache.rs``, ``sync.rs``, etc.) to use the new centralized getters instead of hardcoded paths or inline `std::env::var` checks.


bsirak:
I kept `UVR_R_VERSIONS_DIR` because that's what we discussed, but I must say it bothers me it's not `UVR_R_INSTALL_DIR` for reasons...

nbafrank:
Hey @bsirak ! Thanks so much for this!  You are totally right about `UVR_R_INSTALL_DIR` because it is more standardized with both `uv` but also pairs with `UVR_INSTALL_DIR`, go for it and make that change!

I have a few things for you!

**Naming — you're right!**                                                                                                                
                                                         
  Let's rename UVR_R_VERSIONS_DIR → UVR_R_INSTALL_DIR as discussed above!                                         
                                                              
  **CI — formatting only**                                                                                                                 
                                                         
  CI is failing on cargo fmt --all -- --check across all three platforms. No compile or test failures — just rustfmt. Running cargo fmt
   --all locally and pushing should clear it. Diffs are in cache.rs:8, doctor.rs:324, sync.rs:308/:675, and three spots in
  self_update.rs (trailing whitespace, anyhow::bail! wrapping, inline-comment placement).                                              
                                                         
  **Blocking — correctness**

  crates/uvr-core/src/registry/cran.rs:346 — silent CRAN cache invalidation. The .join("cran") moves the CRAN package-index from       
  ~/.uvr/cache/{key}-packages.txt to ~/.uvr/cache/cran/{key}-packages.txt. Every existing user's CRAN cache becomes orphaned on first
  run after upgrade, and the layout is now inconsistent with P3M which stays flat at ~/.uvr/cache/p3m-*. Either drop the .join("cran") 
  here, or namespace P3M too — but pick one. I'd drop it; cache organization can be a separate PR if we want to do it intentionally for
   both registries.

  crates/uvr/src/commands/self_update.rs — UVR_INSTALL_DIR footgun. When UVR_INSTALL_DIR differs from current_exe.parent(), the new    
  binary is written to the override dir but the running binary at current_exe is never touched. Depending on PATH order, the next
  invocation may hit the stale binary and the user thinks the update failed. Either reject the mismatch with a clear error             
  ("UVR_INSTALL_DIR points to X but uvr is running from Y; remove the env var or move uvr to the override dir first"), or at minimum
  emit a warn explaining the user must update PATH.

  self_update.rs — lost error. The old code errored if current_exe.parent() was None; the new code silently falls back to              
  Path::new("."). Please restore the explicit error — silent install-into-CWD is worse than a clear failure.
                                                                                                                                       
  **Polish**                                                 

  - config.rs:32-37 — install_dir() has a TODO-style comment ("We could fall back to ~/.local/bin but if we are just reading an env var
   to override... maybe it just defaults to None... Let's provide a function that returns the env var or None.") — this reads like
  working notes. Either commit to the choice with a one-line rationale, or remove.                                                     
  - run.rs minor — let extra = extra.trim().to_string(); introduces an unnecessary allocation. The original &str worked fine;
  push_str(extra.trim()) would too.                                                                                                    
  - Trailing whitespace in several config.rs doc comments (e.g. /// UVR_CACHE_DIR\n/// ) — rustfmt won't catch these but they're easy
  to clean.                                                                                                                            
                                                         
  **Tests**                                                                                                                                
                                                         
  Could you add a small mod tests in config.rs covering: (1) defaults when env vars are unset, (2) override when set, (3) empty-string 
  env vars falling through to default? Seven new public entry points with zero coverage is the main thing keeping this from being a
  clean merge.                                                                                                                         
                                                         
  **Out of scope (Bonus!!!) **

  uvr doctor should ideally print the effective values of UVR_CACHE_DIR / UVR_R_INSTALL_DIR / UVR_INSTALL_DIR so users can debug "where
   did my files go". Happy for that to be a separate PR if you'd rather not expand scope here.
   
   
More importantly-> Thank you so much for your help and work here!

bsirak:
@nbafrank, thank you for the info! I'll make the suggested changes, clean it up, and try to get to adding the output to `uvr doctor`.

Do you have an opinion on `UVR_INSTALL_DIR` mismatch handling (error vs warn)? My gut says `UVR_INSTALL_DIR` should only be used by the installer.sh and the self-update should strictly replace the in current location. If you're OK with it, I want to restore the `self_update.rs` to before I touched it and allow the installer.sh to continue to use `UVR_INSTALL_DIR` alone. I know that isn't what we originally discussed, so I'm amenable to anything, but I believe this is current best practice for rust self_update.

nbafrank:
Go for it regarding `UVR_INSTALL_DIR` I am new to many of these practices so I welcome your input and feedback!

bsirak:
hey, sorry. i should've mentioned in my reply that i was just pulling the changes you made into the pr branch. Nothing should've changed yet on my end. I'll post when I think it's ready for another pass.

And no worries, I'm new to most of this too. Certainly on anything this scale.

nbafrank:
No worries. Please write here when ready! Also there are some failing CI tests so check on those as well. Thanks for the help!

bsirak:
@nbafrank
ok, I've updated with a first pass on fixes. I'm definitely not done yet, but I think I addressed everything. Couple of things:
1. the testing in `env_vars.rs` needs a rework. I believe everything will pass right now, but if it ever fails it could go very bad. I'll look at it again tomorrow.
2. I reworked and left `UVR_INSTALL_DIR` in the `env_var.rs` crate for now, but it's not used anymore except for the `install.sh` file. I can remove it if you'd like, but i figured it would be nice to have in the list.
3. I noticed in the `cran.rs` and `p3m.rs` crates the cache directory is falling back to the `cwd` with `.unwrap_or_else(|| PathBuf::from("."))`. Are we ok with this or should we error when both UVR_CACHE_DIR isn't set and the home directory can't be found?

if you want to take a quick look and let me know if you have any other comments or suggestions, I'd greatly appreciate it. Otherwise, I'll take a second pass at the testing for `env_vars.rs` tomorrow evening and hopefully remove all unsafe `env::set_var` calls completely.

**5/8 Update**: fixed the comment indent clippy didn't like, still need to redo env var testing...

nbafrank:
v2 looks great — every blocker from the first review is addressed, plus you snuck in the doctor.rs visibility and the shared \`read_env_var\` helper that I'd flagged as out-of-scope-but-nice. Tests pass on all three platforms now. Two real items still to fix before merge, plus two minor nits:

## Blockers

**B1. Test is panic-unsafe** — \`crates/uvr-core/src/env_vars.rs:108-185\`

The backup/restore pattern is sound under normal execution but not under panic. If any \`assert!\` / \`assert_eq!\` fires before line 178, the test runner unwinds and the restore loop never runs. Every cleared/mutated env var leaks into whatever test runs next in the same process. The comment at line 97 promises "don't permanently mess up the test runner environment" — the code can't keep that promise as written.

Standard fix is a RAII guard:

\`\`\`rust
struct EnvGuard { backups: HashMap<&'static str, Option<String>> }

impl EnvGuard {
    fn new(vars: &[&'static str]) -> Self {
        let mut backups = HashMap::new();
        for &v in vars {
            backups.insert(v, env::var(v).ok());
            env::remove_var(v);
        }
        Self { backups }
    }
}

impl Drop for EnvGuard {
    fn drop(&mut self) {
        for (var, val) in &self.backups {
            match val {
                Some(v) => env::set_var(var, v),
                None => env::remove_var(var),
            }
        }
    }
}
\`\`\`

Then \`let _guard = EnvGuard::new(&vars_to_test);\` at the top and the rest of the test body works as-is. \`Drop\` runs on panic. ~20 lines.

**B2. \`doctor.rs::check_env_vars()\` uses always-green glyph** — \`crates/uvr/src/commands/doctor.rs\`

\`ui::check(true, name, msg, LABEL_W)\` renders a green ✓ unconditionally. So:
- \`UVR_R_INSTALL_DIR=/nonexistent/path\` → green ✓ (looks valid; isn't)
- \`UVR_INSTALL_DIR\` not set → green ✓ next to "not set" (reads as "congratulations, it's not set")

The first case actively misleads users — the whole point of \`uvr doctor\` is to surface misconfiguration. Two clean options:

1. **Validate**: pass \`ok = path.exists()\` (or \`true\` when None for "not set" rows). Nonexistent paths render with a red X.
2. **Drop the pass/fail semantic**: switch to \`ui::bullet_dim\` for pure informational display without a glyph at all.

I'd lean (2) for the "not set" case and (1) for the set-to-a-path case — i.e., when the value is None, just print info-style; when set, validate.

## Nits (good to land but not blocking)

**N1. \`UVR_INSTALL_DIR\` docstring is aspirational** — \`env_vars.rs:36-41\`

The doc says "Used primarily by the \`uvr self-update\` command to determine where to place the newly downloaded executable." But \`self_update.rs\` doesn't call \`env_vars::install_dir()\` — it uses \`std::env::current_exe().parent()\`. So a user reading the docs and setting \`UVR_INSTALL_DIR\` will be silently ignored.

Either tweak the docstring to "Currently informational; self-update does not yet honour this variable — see issue #38 / #79 for plans" or skip exposing it for now and add it back with the actual self-update integration. 30-second docstring fix is fine.

**N2. Dead \`base\` in \`cache.rs:8\`**

\`\`\`rust
let base = dirs::home_dir().unwrap_or_else(|| PathBuf::from(".")).join(".uvr");
let cache_dir = uvr_core::env_vars::cache_dir().unwrap_or_else(|| base.join("cache"));
\`\`\`

\`env_vars::cache_dir()\` already encapsulates the home-dir-based default. \`base\` is allocated only to be used as the unwrap fallback — when \`UVR_CACHE_DIR\` is set, \`base\` is discarded. Cleaner:

\`\`\`rust
let cache_dir = uvr_core::env_vars::cache_dir().unwrap_or_else(|| PathBuf::from("."));
\`\`\`

## Once those two land

I'll merge.

Real appreciation for the iteration speed and quality of the v2 fixes — the consolidated test design (single test to avoid env-var race) was exactly the right call, and the doctor.rs visibility is a thoughtful bonus.

nbafrank:
p.s.
yeaaaah the Rust formatting is a drag... I fail the tests so often now it's not even funny...

bsirak:
ok, updated with suggestions & corrections. I'm not great with testing, so if I need to take another pass on that then let me know. I just implemented you suggestion (I think).

PS. I feel like I've run `fmt --all` and `clippy -- -D warnings` 8 million times, so if there's some formatting issues then I quit...

nbafrank:
Thank you @bsirak !!! We need more contributions like yours :)

bsirak:
Happy to contribute! I'll keep looking for places I might be able to help.

Issue #31, opened 2026-04-15
bug: uvr ignores custom repositories


pat-s:
The image here has `https://cran.rpkgs.com/arm64/alpine323/latest` set as the "CRAN" repos option. However, `uvr` ignores this and always uses `https://cran.r-project.org/src/contrib/` instead, which only serves source packages. 

```sh
docker run --rm devxygmbh/r-alpine:4-3.23 sh -c '
set -ex
curl -fsSL https://raw.githubusercontent.com/nbafrank/uvr/main/install.sh | UVR_INSTALL_DIR=/usr/local/bin sh
mkdir /tmp/testpkg && cd /tmp/testpkg
cat > DESCRIPTION <<EOF
Package: testpkg
Title: Test
Version: 0.0.1
Imports: xml2
EOF
uvr init
uvr lock
uvr sync
'
```

It would also be great if the repos in use would be logged during install. See my `FROM $REPO` suggestions in the snippet below. 
You might also want to think about how this works if multiple repos are configured.

```
+ uvr lock
 INFO Downloading cran PACKAGES.gz FROM $REPO...
 INFO cran index: 23594 packages ($REPO)
✓ Lockfile updated (3 packages)
+ uvr sync
→ Installing 3 package(s)...
```

nbafrank:
Fixed in 3959ad8. Custom repositories defined in `[[sources]]` are now checked **before** CRAN in the registry chain, so packages from r-universe or private repos take priority.

Previously the chain was CRAN → custom → Bioconductor, meaning CRAN always won if a package existed in both. Now it's custom → CRAN → Bioconductor.

pat-s:
> always won if a package existed in both

That's an issue per se, at least for `install.packages()` and should be avoided, if possible. 
`uvr` might be better at conflict resolution (hopefully), so this might be less of an issue.

My issue though was about custom repos being ignored completely - I'll take is as fixed and will try it out in the next release!

nbafrank:
Yes please, test it and tell me if it works. If it does, close this issue and if not I will try and fix it!

nbafrank:
@pat-s — the precedence fix landed in v0.2.9 (custom `[[sources]]` repos checked before CRAN, then Bioconductor). Have you had a chance to retry on a tagged release? If it works for your r-multiverse / private-repo case, OK to close. If not, repro recipe + I'll have another look.

pat-s:
[...]

Issue #30, opened 2026-04-15
bug: syslib installation under Alpine doesn't work


pat-s:
The following should advertise to install `libxml2` for {xml2} to work but nothing happens.
The rule [exists](https://github.com/rstudio/r-system-requirements/blob/main/rules/libxml2.json).

v0.2.6

```sh
docker run --rm devxygmbh/r-alpine:4-3.23 sh -c '
set -ex
curl -fsSL https://raw.githubusercontent.com/nbafrank/uvr/main/install.sh | UVR_INSTALL_DIR=/usr/local/bin sh
mkdir /tmp/testpkg && cd /tmp/testpkg
cat > DESCRIPTION <<EOF
Package: testpkg
Title: Test
Version: 0.0.1
Imports: xml2
EOF
uvr init
uvr lock
uvr sync
'
```

```
+ uvr sync
→ Installing 3 package(s)...
error: Failed to install xml2: R CMD INSTALL failed for 'xml2' (exit 1):
uvr: linking library at /tmp/testpkg/.uvr/library
* installing *source* package ‘xml2’ ...
** this is package ‘xml2’ version ‘1.5.2’
** package ‘xml2’ successfully unpacked and MD5 sums checked
** using non-staged installation
ERROR: configuration failed for package ‘xml2’
* removing ‘/tmp/testpkg/.uvr/library/xml2’
```

nbafrank:
Added Alpine/`apk` support to system dependency detection. `filter_missing()` now checks `apk info -e` when `dpkg`/`rpm` aren't available, and the install command suggestion uses `apk add` on Alpine instead of `apt-get install`. In latest main.

pat-s:
*AI generated comment*

---

Thanks for adding `apk` support to `filter_missing()` — but the underlying issue is still present. I dug into it and the root cause is upstream of that fix:

## Root cause

`resolve_system_deps()` in `sysreqs.rs` queries the **Posit Package Manager API**:

```
https://packagemanager.posit.co/__api__/repos/1/sysreqs?all=false&pkgname=fs&distribution=alpine&release=3.21
```

This API returns `{"code":14,"error":"Unsupported system"}` for Alpine (all releases I tested: 3.15–3.21). uvr then silently returns an empty list because the non-success status branch just does `return Ok(vec![])`.

So `filter_missing()` never even runs — there are zero resolved deps to filter.

## Why it's confusing

[...]

PR #88, opened 2026-05-11
fix: alpine binary install + custom binary sources via [[sources]] / R repos / UVR_REPOS

pat-s:
@nbafrank See https://ci.codefloe.com/repos/148/pipeline/390/12 for a live example of how these PR changes may work.

## Problem

Reported as failing on `r-alpine:4.5-3.23`:

```
> Installing 83 package(s): 80 binary - 3 from source
  Failed to install rlang
  Failed to extract 'rlang/DESCRIPTION' from 'rlang': failed to unpack `rlang/DESCRIPTION` …
```

Three independent issues stacked into one symptom:

1. **`detect_posit_distro_slug()` silently rewrote alpine to `ubuntu-2204`** (`r_version/downloader.rs:213-230`'s `_ => "ubuntu-2204"` fallthrough). Result: alpine downloaded Ubuntu Jammy glibc binaries from P3M. Even if those had extracted, the `.so` files would never have loaded on musl.

2. **`[[sources]]` could not supply binaries.** `sync.rs` unconditionally consulted P3M and preferred its URLs over whatever the lock-time `RegistryChain` resolved — so even with a `[[sources]]` entry pointing at `cran.rpkgs.com`, uvr ignored it for binaries.

3. **`extract_tgz` used `tar::Entry::unpack`**, which preserves mtime/permissions/ownership and uses `OpenOptions::create_new` + a remove-then-recreate dance. Some CI filesystems (Drone runners, overlayfs, FUSE mounts) reject specific syscalls in that chain; the failure was the actual error at `rlang/DESCRIPTION`, not the libc mismatch.

#30 (alpine sysreqs) and #31 (custom repos before CRAN in lock-time chain) addressed sysreqs and lock-time resolution. The binary download path was still unfixed.

## What this PR ships

**Distro detection fix.** Alpine maps to `alpine-X.Y`, which `ppm_linux_codename` cannot translate → `P3MBinaryIndex` returns empty → sync cleanly falls through to source. No more silent jammy binaries on musl.

**Custom binary sources via `[[sources]]`.** Auto-detection: an entry's `Built:` field is matched against the running host. At least one host-matching source ⇒ P3M is suppressed and custom sources are queried in declaration order. Source-only customs (r-multiverse, r-universe) coexist with P3M as today. Path traversal hardened. Vendor segment of the platform triple is ignored when matching (alpine R reports `aarch64-unknown-linux-musl`, Ubuntu R reports `aarch64-pc-linux-musl`).

**Tarball-sniff fallback for repos that don't advertise `Built:` in PACKAGES.** [cran.rpkgs.com](https://cran.rpkgs.com) intentionally publishes a unified source-style PACKAGES so non-binary-aware tooling still works. Post-download uvr peeks at the tarball's DESCRIPTION; if `Built:` is there and matches the host, the install is reclassified as binary and routed through the pure-Rust `install_binary_package` fast path. R itself already does this — uvr now matches.

**Source auto-discovery from R.** If R has `options(repos)` set (typically via `Rprofile.site`), uvr picks up non-CRAN entries automatically. No `UVR_REPOS` or `[[sources]]` needed when R already knows where the mirror is. Same way pak/renv work. Filters `@CRAN@`, `r-project.org`, `bioconductor.org`.

**`UVR_REPOS` env var.** Comma-separated URLs become `[[sources]]` entries at runtime, names auto-derived from URL host. Removes the need for ad-hoc `printf >> uvr.toml` in CI.

**`extract_tgz` rewritten** with explicit `fs::create_dir_all` + `fs::File::create` + `io::copy`. No metadata preservation (R packages have no meaningful mtime), no `create_new`, no remove-then-recreate. Symlinks/hardlinks/devices skipped (R packages don't use them). Error messages now include `io::Error.kind()` so any future filesystem failure self-diagnoses. Verified by an integration test that pulls real cran.rpkgs.com tarballs (rlang + jsonlite) and extracts them — runs against both x86_64 and aarch64 alpine 3.23.

**Install summary uses runtime classification.** Both the upfront and final summary now reflect what actually happened (binary fast-path vs `R CMD INSTALL`) rather than the lock-time pre-estimate. The 'No binary repo for X on R Y' info hint only fires when zero packages were binary.

**User-Agent matches real R.** Construction follows `getOption("HTTPUserAgent")`: `R (<ver> <triple> <arch> <os>-<abi>)`. Satisfies PPM's gating and gives cran.rpkgs.com the platform-triple substring it routes on. Sent on `[[sources]]` PACKAGES fetches and binary tarball downloads (was previously hardcoded to `linux-gnu` and only for `/__linux__/` URLs).

## Behavior summary

| Scenario | Before | After |
|---|---|---|
| Alpine, no custom source | Silent jammy glibc download → tar fails | Falls cleanly to source compile; one info line explains why |
| Alpine + `[[sources]]` = cran.rpkgs.com | `[[sources]]` ignored at sync time → P3M wins → wrong-libc | Sources consulted, tarball-sniff reclassifies as binary, fast-path install |
| Alpine + `Rprofile.site` has `cran.rpkgs.com` | uvr ignored R's config | Auto-discovered at lock/sync time |
| Ubuntu, no custom source | P3M binaries | P3M binaries (unchanged) |
| Drone CI / overlayfs / FUSE | `extract_tgz` failed at first entry, no useful error | Extracts cleanly; if it ever fails again, error includes `io::Error.kind()` |

## Closes

This is a separate concern from #30. References #31 (which gave lock-time priority to `[[sources]]`; this PR extends that priority into the sync-time binary URL lookup as well).


nbafrank:
Hi @pat-s ! Thanks for much for your work! Apologies for the delay in looking at this.The alpine fix, BuiltInfo matching, and extract_tgz rewrite are all clean. Three things to sort before merge:

 1. Windows CI is failing because of an IPv6/IPv4 loopback race against the mock server, not your logic. Easiest fix is #[cfg(not(target_os = "windows"))] on lock_with_binary_capable_source_records_source_urls — the parser tests cover that path on  Windows already.  
                                       
  2. discover_r_repos reverses #31's decision about not reading R session state. It also has a concrete bug: the --vanilla --slave call suppresses .Rprofile/Rprofile.site loading, which is where users would set their custom repos in the first place. So in vanilla it  can't do what it advertises; in non-vanilla it brings back the lockfile-reproducibility hole. I'd drop it from this PR — UVR_REPOS already covers the CI case.
                                                                                                           
  3. UVR_REPOS is read at lock time, which means two devs with different env vars resolve different uvr.lock from the same uvr.toml. Same reproducibility problem. I'd keep UVR_REPOS at sync time only. Thoughts?

pat-s:
Thanks for the review — all three points are right and I've addressed them in 62359b0:

1. **Windows CI** — added `#[cfg(not(target_os = "windows"))]` on `lock_with_binary_capable_source_records_source_urls` (and on `spawn_rpkgs_stub` so we don't have a dead helper on Windows). Parser tests in `registry::cran` still cover the path on all platforms.

2. **`discover_r_repos` dropped entirely.** You caught a real bug — `R --vanilla` is `--no-save --no-restore --no-init-file --no-site-file --no-environ`, so `Rprofile.site`/`.Rprofile` never loads, meaning `getOption("repos")` returns the placeholder default. The function literally couldn't do what its docstring claimed. Verified locally — setting `options(repos = ...)` in `/etc/R/Rprofile.site` and then calling `R --vanilla -e 'getOption("repos")'` returns `@CRAN@`. And as you said, even if I fixed `--vanilla`, reading R session state at lock time would reintroduce the reproducibility hole #31 ruled out. `UVR_REPOS` (now sync-only) and `[[sources]]` cover all the real use cases.

3. **`UVR_REPOS` moved to sync-time only.** `lock.rs` now reads only `project.manifest.sources` — env-injected repos are only consulted in `sync.rs` for binary URL upgrades. Same shape as how P3M URLs are added at sync without touching the lockfile. Two devs with different `UVR_REPOS` produce the same `uvr.lock` from the same `uvr.toml`.

CHANGELOG entries cleaned up. 233 tests pass, clippy + fmt clean. Ready for another look.

nbafrank:
@pat-s almost there! Just need to fix the macos test (likely a rust formatting issue)

pat-s:
Pushed `942ee2f` to fix the macOS CI failure in `lock_with_binary_capable_source_records_source_urls`.

Different bug from the Windows one. The stub server's `TcpListener::accept()` returns a socket that inherits the listener's non-blocking flag on macOS — so `socket.read()` returns `WouldBlock` immediately, our `.unwrap_or(0)` gives an empty request, we 404 it and close before reqwest finishes its GET. reqwest surfaces that as `received unexpected message from connection`.

Fix:
- Force the accepted socket back to blocking and attach a 5s read timeout (never hang on malformed requests).
- Send `Connection: close` so reqwest doesn't keep-alive against our one-shot handler.
- Half-close the write side after the response.

Verified locally — test passes 5/5 consecutive runs.

nbafrank:
@pat-s we have now multiple tests failing 🤔for some reason

PR #101, opened 2026-05-26
feat: Forgejo remote support

pat-s:
## Summary

Adds support for installing R packages from Forgejo instances via a new `forgejo::` prefix.

```sh
uvr add forgejo::codefloe.com/pat-s/mypkg@v0.1.0
```

- New `forgejo::host/owner/repo[@ref]` syntax works in three places: `uvr add` CLI, `uvr.toml`'s `git = "..."` field, and DESCRIPTION `Remotes:` entries.
- New `PackageSource::Forgejo { host }` lockfile variant serialized as `forgejo:<host>`.
- Forgejo API endpoints (`/api/v1/repos/{o}/{r}/{commits,raw,archive}`) used end-to-end — no Gitea-only routes.
- Per-host auth via `UVR_FORGEJO_TOKEN_<HOST>` (or single `UVR_FORGEJO_TOKEN`), sent as `Authorization: token <value>`. Mirrors the recent `GITHUB_TOKEN` work for github.
- Lock-time BFS generalized to `resolve_git_deps`, dispatching per-prefix. Cross-host `Remotes:` chains (github↔forgejo) walk transitively.
- `uvr export` emits Forgejo packages as `Source: Git`, `RemoteType: git2r`, `RemoteUrl: https://host/owner/repo` so they can be re-imported by renv.

## Design + plan

Design doc and the 12-task implementation plan are committed in `docs/superpowers/`:

- `docs/superpowers/specs/2026-05-26-forgejo-remote-support-design.md`
- `docs/superpowers/plans/2026-05-26-forgejo-remote-support.md`

## Notes for reviewers

- The integration test at `crates/uvr-core/tests/forgejo_live.rs` is `#[ignore]`d by default. Its owner/repo are placeholders (`codeberg.org/Codeberg/Documentation`) — swap in a real public Forgejo-hosted R package before relying on it.
- No new crate dependencies. Uses `serde_json` for the commits-endpoint JSON parse; everything else mirrors `github.rs`.
- All 347 lib/integration tests pass; `cargo fmt --check` clean.

### DIFF
--- a/crates/uvr-core/src/installer/download.rs
+++ b/crates/uvr-core/src/installer/download.rs
@@ -50,6 +50,7 @@ impl Downloader {
                 let fallback_url = spec.fallback_url.map(str::to_string);
                 let is_binary = spec.is_binary;
                 let user_agent = spec.user_agent.map(str::to_string);
+                let auth_header = spec.auth_header.map(str::to_string);
                 // Binary packages: lockfile checksum is for the source tarball, not the
                 // P3M binary. Skip verification on binary downloads, but keep the
                 // checksum for the fallback path which downloads the source tarball.
@@ -73,6 +74,7 @@ impl Downloader {
                         &url,
                         primary_checksum.as_deref(),
                         user_agent.as_deref(),
+                        auth_header.as_deref(),
                         &mp,
                     )
                     .await;
@@ -104,6 +106,7 @@ impl Downloader {
                                 fallback,
                                 source_checksum.as_deref(),
                                 None, // fallback URL is plain CRAN source — no UA override needed
+                                None, // fallback is a different host; never forward primary auth
                                 &mp,
                             )
                             .await?;
@@ -138,6 +141,11 @@ pub struct DownloadSpec<'a> {
     /// served at the same URL, and the default `uvr/x.y.z` UA gets you
     /// source. None = use the client's default UA.
     pub user_agent: Option<&'a str>,
+    /// Optional `Authorization` header value (e.g. `"token <forgejo>"` or
+    /// `"Bearer <github>"`). Forwarded to the primary URL only. Fallback
+    /// URLs (CRAN Archive, P3M → source) deliberately drop the header —
+    /// the token is registry-scoped and shouldn't leak to other hosts.
+    pub auth_header: Option<&'a str>,
 }
 
 /// Result of downloading a single package.
@@ -176,6 +184,7 @@ async fn download_one(
     url: &str,
     expected_checksum: Option<&str>,
     user_agent: Option<&str>,
+    auth_header: Option<&str>,
     mp: &MultiProgress,
 ) -> Result<PathBuf> {
     // Derive the cache filename from the URL so that source tarballs (.tar.gz)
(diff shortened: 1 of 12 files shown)


nbafrank:
Thanks man!!! Will review soon @pat-s !

nbafrank:
Hey @pat-s, finally got to a proper read-through. Few things I'd like resolved before we merge:

**1. The `fetch_commit_sha` endpoint 404s.** `GET /api/v1/repos/{owner}/{repo}/commits/{ref}` doesn't exist on Forgejo — I curl-checked against codefloe.com (15.0.2) and codeberg.org (15.0.0), both return 404. Looks like the ignored `forgejo_live.rs` is the only thing that would've caught it, and it never ran. The working form is `commits?sha={ref}&limit=1` (returns a single-element array, read `[0].sha`). `forgejo.rs:181`.

**2. `uvr sync` doesn't forward the token to the tarball download.** `DownloadSpec` has no auth field and `sync.rs:67` builds `PkgPlan` without consulting `PackageSource::Forgejo { host }`. So `uvr add forgejo::private-host/...` would lock fine and then silently 401 on install. Needs an `auth_header: Option<String>` on `DownloadSpec` populated via `forgejo_token(host)` in sync.

**3.** Design doc says `RemoteType: gitea`, code emits `git2r` (commit `c4c1b3e`). What flipped? Mind updating the doc with the reason — don't want to leave a contradiction in the committed spec.

Rest of the diff looks good — lockfile round-trip, BFS cycle detection, token redaction in errors, the `forgejo_token` env normalization. A few smaller things I'll let you decide on (port validation accepts non-numeric, the live test's placeholder owner/repo was never replaced, three slightly-different spec parsers in add/lock/manifest) — happy to file them as follow-ups if you want to keep this PR scoped.

pat-s:
Fixed all three. Verified each before implementing.

**1. `commits/{ref}` 404 — confirmed.** curl-tested codeberg.org myself, `/commits/main` returns 404, `/commits?sha=main&limit=1` returns a JSON array. `fetch_commit_sha` now uses the list-commits form and deserializes `Vec<CommitObj>`, returning an error if the array is empty (ref doesn't exist). Spec doc table + the "Differences from github.rs" paragraph updated to match. `651ec38`, `55bda72`.

**2. sync token forwarding — confirmed missing.** Added `auth_header: Option<&'a str>` to `DownloadSpec`, plumbed through `download_one`. sync.rs precomputes a `Vec<Option<String>>` of `"token <tok>"` headers per plan via `forgejo_token(host)`, zips with plans, and sets `auth_header` on each spec. Fallback URLs explicitly pass `None` so the registry-scoped token never leaks to a different host (e.g. CRAN Archive fallback). `e189c92`.

**3. `RemoteType: gitea` vs `git2r` — spec doc was wrong.** renv has no `gitea` source type; `git2r` is the closest renv-understood `RemoteType` for a generic clone-from-URL flow. Spec doc updated with the reasoning inline (touched-files table). `fc80f96`. While I was in the doc I also fixed the auth-normalization prose ("lowercase" → "uppercase") so it matches the code. `55bda72`.

Skipping the three smaller items from your last paragraph (non-numeric port validation, placeholder owner/repo in `forgejo_live.rs`, three slightly-different spec parsers) — happy to take them as follow-ups so this PR stays scoped to the three blockers.

`cargo clippy --all-targets -- -D warnings` clean, all 347 tests pass.

nbafrank:
Nice — all three verified locally. Clippy clean, 347 tests pass, endpoint fix returns sane error on a missing ref, fallback path correctly drops the token. Merging.

PR #115, opened 2026-06-13
enh: support `--install-system-deps` flag in `uvr add`

pat-s:
## Summary
- Add `--install-system-deps` to `uvr add`, mirroring the existing flag on `uvr sync`.
- Syslib (sysreqs) detection already runs during `uvr add` since it installs via `sync::install_from_lockfile`, but the opt-in to auto-install detected system libraries through the platform package manager (`apk add` / `apt-get install` / `dnf install`) was only exposed on `sync`.
- The `Add` handler now sets `UVR_INSTALL_SYSREQS=1` when the flag is passed, the same mechanism the `Sync` handler uses, so no bool threading through the shared install path.

### DIFF
--- a/crates/uvr/src/cli.rs
+++ b/crates/uvr/src/cli.rs
@@ -156,6 +156,15 @@ pub struct AddArgs {
     /// there's no fresh lockfile to install from.) (#76)
     #[arg(long)]
     pub no_install: bool,
+
+    /// When missing system libraries are detected, run the platform's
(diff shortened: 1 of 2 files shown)


nbafrank:
Thanks, merged! Nice to close it! `add` and `sync` sharing the `UVR_INSTALL_SYSREQS` mechanism is exactly right. Ships in the next release.

Issue #123, opened 2026-06-23
Add support for gitlab


pteridin:
Hello everyone,

thank you for this useful package. Would it be possible to add support for packages hosted on gitlab? This would facilitate using in-house packages that are not available publicly.

All the best
Tobias

gdevenyi:
Concrete blocker for `uvr add gitlab::` support in `add.rs:37-50`:

- `gitlab::` prefix is not handled (only `forgejo::` at `:19-34`)
- Bare `gitlab.com/user/repo` falls into the GitHub heuristic at `:37`, fails the `parts.len() != 2` check at `:46` (3 path segments), errors with "Invalid GitHub spec"

Adding `gitlab::` parsing parallel to the `forgejo::` branch would unblock this.

Audit context: comprehensive codebase review.

PR #173, opened 2026-07-23
Design proposal: uv feature parity (R-native lens)

gdevenyi:
## What this is

A design spec auditing uvr against [`uv`](https://github.com/astral-sh/uv) and proposing which uv capabilities are worth bringing to uvr — judged under an **R-native workflow-parity** lens: include a uv feature only if a real R developer needs that *outcome*, then reshape it to R idioms; drop Python-ecosystem artifacts as category errors.

It follows the existing `docs/superpowers/specs/…-design.md` convention. **Design only — no code changes.** Forgejo-grade, TDD implementation plans (one per feature, under `docs/superpowers/plans/`) follow once the design is approved.

## Six features proposed (three tiers)

**P0**
- **Shell activation** — `source .uvr/activate` puts the project's managed R + isolated library on your shell (the `source .venv/bin/activate` ergonomic), via a sourceable shim that delegates to a dynamic `uvr activate --emit <shell>`, so it can't go stale on an R-version change.
- **Inline per-script dependencies** — a PEP 723-style `# /// script` header so `uvr run script.R` runs standalone in any directory and reproduces anywhere (the live edge over `rv`). Reuses the existing `--with` ephemeral-env cache.
- **Private-repo auth** — token/basic-auth + `~/.netrc` for authenticated repos (private Posit Package Manager / internal mirror), unblocking the corporate/university audience. No secrets in `uvr.toml`.

**P1**
- **Dependency sources** — local `path`, arbitrary git host (GitLab / Bitbucket / generic — addresses #123), and direct URL tarball. Generalizes the existing Forgejo prefix scheme.
- **Ergonomic parity** — `cache prune/dir/size`, `r dir/find`.

**P2**
- **Resolution controls** — `--resolution lowest`, `--exclude-newer` (via dated PPM snapshot), overrides/constraints, `lock --upgrade-package`.

## Deliberately excluded (category errors under the lens)

`uv build`/`publish` (devtools + CRAN own this) · `uvx`/`uv tool` (R packages don't ship CLIs) · `uv pip` (breaks lockfile-first) · `uv venv` create-primitive (activation covers the real need) · dependency groups + extras · editable/workspaces · `--prerelease` (no CRAN prerelease channel) · keyring / `uv auth` CLI · `uv version` · export-to-DESCRIPTION (R-native, no uv analog).

## Already at parity (no work)

- **Universal lockfile** — `uvr.lock` already pins source, not OS/arch; the binary is chosen per-host at sync, so one lockfile serves every platform.
- **Targeted upgrade** — `uvr update <pkg>` already re-resolves only the named package.

## Feedback wanted

The design calls out four debatable engineering decisions — the inline-script header format + auto-provisioning R by default, growing `PackageSource` for the new source kinds, implementing `--exclude-newer` as a snapshot-URL redirect, and the `auth` discriminator on `[[sources]]`. Happy to adjust any before the plans are written.

See `docs/superpowers/specs/2026-07-23-uv-feature-parity-design.md` for the full parity matrix and per-feature designs.

### DIFF
--- a/docs/design/2026-07-23-uv-feature-parity-design.md
+++ b/docs/design/2026-07-23-uv-feature-parity-design.md
@@ -1,6 +1,6 @@
 # uv feature parity — design
 
-Status: draft — awaiting approval
+Status: approved 2026-07-27 (@nbafrank) — implementation plans follow, one PR per feature
 Author: gdevenyi
 Date: 2026-07-23
 
@@ -8,9 +8,9 @@ Date: 2026-07-23
 
 Audit uvr's feature set against [`uv`](https://github.com/astral-sh/uv) and close the
 gaps that a real R developer's workflow needs — reshaped to R idioms rather than
-copied literally. This document is the umbrella design; each accepted feature gets a
-companion forgejo-grade implementation plan under
-`docs/superpowers/plans/2026-07-23-uv-parity-NN-*.md`.
+copied literally. This document is the umbrella design; each accepted feature then gets
+its own forgejo-grade implementation plan (`docs/design/2026-07-23-uv-parity-NN-*.md`),
+submitted **one PR per feature** rather than as a pre-approved batch.
 
 Six features are in scope, in three priority tiers:
 
@@ -19,7 +19,7 @@ Six features are in scope, in three priority tiers:
 | P0 | F0 | Shell activation | `source .venv/bin/activate` |
 | P0 | F1 | Inline per-script dependencies | PEP 723 `# /// script` |
 | P0 | F5 | Private-repo authentication | `uv auth` / index credentials |
-| P1 | F3 | Dependency sources (path / git-host / url) | `[tool.uv.sources]` |
+| P1 | F3 | Dependency sources (path / generic git / url) | `[tool.uv.sources]` |
 | P1 | F6 | Ergonomic parity commands | `cache prune/dir/size`, `python dir/find` |
 | P2 | F4 | Resolution controls | `--resolution`, `--exclude-newer`, overrides/constraints |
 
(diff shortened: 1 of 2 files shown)


gdevenyi:
@nbafrank I'm all in on getting uvr to be feature complete versus uv.

I'd like your comments on this spec document. Once its approved (and possibly amended), I'll add all the implementation plans to this PR for review, and then I can knock off each one at a time.

nbafrank:
This is an impressive piece of work — the parity matrix alone is worth merging as documentation, and the exclusions are as well-argued as the inclusions. Overall I'm on board with the direction. Tier-by-tier:

- **F0 (activation), F1 (script headers), F5 (auth)**: accepted as designed. F1 especially — that's the differentiator.
- **F3 (dependency sources)**: accepted, and good timing — #174 just landed GitLab support in exactly the `gitlab::` shape you specify here, so F3 can treat that as its first slice already done. Path/url/generic-git remain.
- **F6**: accepted, cheap wins.
- **F4 (resolver knobs)**: accepted in principle, sequenced last as you propose — I want the seams but agree demand is lowest.

Two logistics notes: let's move the doc to `docs/design/` rather than `docs/superpowers/specs/`, and I'll take the per-feature plans one PR at a time rather than pre-approving the whole sequence. Merging this as the umbrella once the path is moved.

gdevenyi:
Thanks @nbafrank — both logistics notes done, and F3 rewritten around #174. Pushed in 5cb181b (git detected the rename at 87% similarity, so the diff reviews cleanly).

**Logistics**
- Moved to `docs/design/2026-07-23-uv-feature-parity-design.md`, and pointed the per-feature plan path at `docs/design/` to match. I left the existing forgejo spec/plan where they are — say the word if you want those relocated too.
- Recorded that plans ship **one PR per feature**, reviewed and merged before the next is written, rather than as a pre-approved batch. Status line updated to approved.

**F3 now builds on #174 instead of re-specifying GitLab.** Remaining slices are path, url, and generic `git::`. While wiring that up I read #174's diff, and two details correct the data model as I'd drafted it:

1. The lockfile variant is spelled **`Gitlab { host }`**, not `GitLab`.
2. GitLab namespaces **nest arbitrarily deep** (`host/group/subgroup/project`), unlike GitHub's and Forgejo's fixed `host/owner/repo`. `parse_gitlab_parts` handles variable depth — so I've written it into the spec that no later `<kind>::` resolver may assume three segments, since that's an easy trap for the generic-`git::` work.

**One correction on the premise, so the sequencing is right:** #174 is still **open, not merged** — `main` is at `5dfdafb` (v0.4.3 prep) with no GitLab commit — and @pteridin's PR body notes *"Extensive testing was not performed (TBD)."* The `gitlab::` shape is exactly as you described, so the design impact is nil; it only changes ordering. F3 now records #174's merge as a precondition and opens by verifying it against a real GitLab project. If you'd rather I fold that verification into a review pass on #174 itself, happy to.

**One scope call worth flagging:** your "remains" list omitted Bitbucket, and I agree — generic `git::` covers Bitbucket functionally, so a dedicated Bitbucket REST resolver is now an explicit non-goal rather than a planned slice. Easy to revert if you meant otherwise.

Ready to merge as the umbrella whenever you are. I'll start the F0 (activation) plan as its own PR once this lands.

nbafrank:
Thanks! Merged!

PR #174, opened 2026-07-27
Added gitlab as an endpoint

### LINKED ISSUE #123: Add support for gitlab

pteridin:
Hello everyone,

Fixes #123 and makes gitlab available as an endpoint for uvr. Extensive testing was not performed (TBD). As such, please test and report any suggestions to this feature branch.

All the best
pteridin

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2359,7 +2359,7 @@ checksum = "06abde3611657adf66d383f00b093d7faecc7fa57071cce2578660c9f1010821"
 
 [[package]]
 name = "uvr"
-version = "0.4.1"
+version = "0.4.2"
 dependencies = [
  "anyhow",
  "assert_cmd",
@@ -2393,7 +2393,7 @@ dependencies = [
 
 [[package]]
 name = "uvr-core"
-version = "0.4.1"
+version = "0.4.2"
 dependencies = [
  "anyhow",
  "chrono",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2359,7 +2359,7 @@ checksum = "06abde3611657adf66d383f00b093d7faecc7fa57071cce2578660c9f1010821"
 
 [[package]]
 name = "uvr"
-version = "0.4.2"
+version = "0.4.1"
 dependencies = [
  "anyhow",
  "assert_cmd",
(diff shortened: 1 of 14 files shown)


gdevenyi:
This is a feature pending in my comprehensive #173 "uv feature parity" proposal.

nbafrank:
This is a genuinely strong first contribution — you clearly read the Forgejo implementation before starting, and it shows: ref validation, per-host tokens, nested-subgroup `%2F` encoding, lockfile fallback, renv export — all correct and well-tested. And the `gitlab::host/group/repo` shape happens to match exactly where we're heading with generalized git sources (#173), so this slots right in.

A few things before merge:
1. `cargo fmt` — that's all CI is unhappy about.
2. Could you rebase onto current main? It merges cleanly (I checked), but please drop the Cargo.toml/lock version bump in the process — versions get set at tag time on this end.
3. You mentioned testing was TBD: could you add a `gitlab_live.rs` mirroring `tests/forgejo_live.rs` (`#[ignore]`d, network-gated) against a small public R package on gitlab.com? This touches the network/install path so I want one real end-to-end run before it ships.
4. Tiny nit: in `resolve_git_pkg_names`, swap the hand-rolled `parts[1..].join("%2F")` for `urlencoding::encode(...)` like gitlab.rs does — same output today, one fewer encoding site to drift.

Also noticed the `combine_logs` improvement riding along in the second commit — that's a good fix in its own right (R really does scatter errors across both streams), just calling it out so it's reviewed on purpose rather than by accident.

Thanks for taking on the longest-standing request in the tracker — this is a careful implementation of it.

nbafrank:
But can't be deployed until checks pass

pteridin:
Thank you!

> Also noticed the combine_logs improvement riding along in the second commit — that's a good fix in its own right (R really does scatter errors across both streams), just calling it out so it's reviewed on purpose rather than by accident.

That was an accident on my part to include it since I ran into some weird R error without helping context along the way. But nevertheless happy to include if it makes sense.

> read the Forgejo implementation

Most of it is copy-paste tbh.

> A few things before merge 

Will prepare the changes whenever my time allows.

PR #197, opened 2026-07-29
F0: shell activation (source .uvr/activate)

### LINKED ISSUE #177: Extract a shared R environment builder
## Parent

One vertical slice of the umbrella design [`docs/design/2026-07-23-uv-feature-parity-design.md`](https://github.com/nbafrank/uvr/blob/main/docs/design/2026-07-23-uv-feature-parity-design.md) (merged in #173).

## What to build

`uvr run` builds the isolated R environment inline: the project library, a shadowed site library, runtime library paths, and suppression of user environment files. Shell activation and inline script headers both need that exact same environment, so this prefactor moves the construction behind one reusable builder in the core crate.

No user-visible behaviour changes. This exists so the two P0 features can't drift apart from `uvr run` or from each other.

## Acceptance criteria

- [ ] The isolated-environment variables come from a single reusable builder in the core crate, not assembled inline in the run command.
- [ ] `uvr run` behaves identically before and after, in both script and interactive mode — same variables, same values, same precedence for `--with` and extra-library overrides.
- [ ] The builder is unit-tested directly, including the `--with` library taking precedence over the project library.
- [ ] Existing `uvr run` tests pass unchanged.

## Blocked by

None — can start immediately.

### LINKED ISSUE #178: Activate a project in the current shell (bash, zsh, sh)
## Parent

One vertical slice of the umbrella design [`docs/design/2026-07-23-uv-feature-parity-design.md`](https://github.com/nbafrank/uvr/blob/main/docs/design/2026-07-23-uv-feature-parity-design.md) (merged in #173).

## What to build

Activate a uvr project in the current shell so a bare `R` or `Rscript` uses the project's managed R version and isolated library, with no `uvr run` prefix — the interactive-console and IDE-terminal workflow.

Sourcing the generated shim delegates back to uvr at activation time rather than baking values into a file, so changing the project's R version can never leave a stale interpreter behind. `deactivate` restores the shell.

## Acceptance criteria

- [ ] `uvr activate` with no arguments prints the single line to source.
- [ ] An emit mode prints shell statements for POSIX shells; a write-shim mode (re)creates the shim, and `uvr init` creates it for new projects.
- [ ] After sourcing, `R` and `Rscript` resolve to the project's managed R and see only the project library.
- [ ] `deactivate` restores the prior PATH and R library variables and removes itself.
- [ ] Changing the project's R version and re-activating picks up the new interpreter with no regeneration step.
- [ ] Activating outside a project fails with a clear message and leaves the shell untouched.
- [ ] The generated shim is git-ignored.

## Blocked by

- #177 — Extract a shared R environment builder

### LINKED ISSUE #179: Activation for fish and PowerShell
## Parent

One vertical slice of the umbrella design [`docs/design/2026-07-23-uv-feature-parity-design.md`](https://github.com/nbafrank/uvr/blob/main/docs/design/2026-07-23-uv-feature-parity-design.md) (merged in #173).

## What to build

The same activation experience in the other two shells uvr already supports for completions, using each shell's own syntax for setting, exporting and erasing variables, and for defining and removing the `deactivate` function.

## Acceptance criteria

- [ ] Emit mode produces valid, sourceable output for fish and for PowerShell.
- [ ] Write-shim mode creates the fish and PowerShell shims alongside the POSIX one.
- [ ] Activation and `deactivate` round-trip correctly in both shells.
- [ ] Emitted values match the shared environment builder exactly, so shells cannot drift apart.

## Blocked by

- #178 — Activate a project in the current shell (bash, zsh, sh)

### LINKED ISSUE #180: Opt-in activated-project prompt
## Parent

One vertical slice of the umbrella design [`docs/design/2026-07-23-uv-feature-parity-design.md`](https://github.com/nbafrank/uvr/blob/main/docs/design/2026-07-23-uv-feature-parity-design.md) (merged in #173).

## What to build

When a project is activated, optionally show its name in the shell prompt, the way virtualenv activation does. Off by default — a mutated prompt is contentious and many users manage their own. Enabled per-project in the manifest or per-shell by environment variable.

## Acceptance criteria

- [ ] The prompt is unchanged by default.
- [ ] Enabling the opt-in prefixes the prompt with the project name.
- [ ] `deactivate` restores the original prompt exactly.
- [ ] Works in every shell that activation supports.

## Blocked by

- #178 — Activate a project in the current shell (bash, zsh, sh)

gdevenyi:
Implements **F0 — Shell activation**, the first feature from the [uv feature-parity design](https://github.com/nbafrank/uvr/blob/main/docs/design/2026-07-23-uv-feature-parity-design.md) merged in #173. One PR per feature, as you asked.

Closes #177, closes #178, closes #179, closes #180.

## What this gives you

```sh
source .uvr/activate      # bash, zsh, sh
source .uvr/activate.fish # fish
. .uvr/activate.ps1       # PowerShell

R                         # uses the project's R and .uvr/library/
deactivate                # restore your shell
```

The interactive-console and IDE-terminal workflow, without prefixing everything with `uvr run`.

## The one design decision worth your attention

**The shims contain no paths.** A Python venv can bake absolute paths into its activate script because the venv *is* the interpreter — it cannot change underneath the file. A uvr environment is computed from `uvr.toml` and `.r-version`, which `uvr r use` / `uvr r pin` edit. A static script would keep pointing at the old R after such an edit.

So `.uvr/activate` is a thin shim that asks the binary to recompute the environment as it is sourced. Staleness becomes structurally impossible rather than something to remember. It also only `eval`s on success, so a failed resolve (no project, no R) leaves your shell untouched instead of half-activated.

## Commits, mapped to the issues

| Commit | Issue |
|---|---|
| `r_env`: extract the isolated R environment into a shared builder | #177 |
| `activate`: `source .uvr/activate` to enter the project shell | #178 |
| `activate`: fish and PowerShell support | #179 |
| `activate`: opt-in activated-project prompt | #180 |

Plus four follow-ups from a two-axis review (standards + spec) run before opening this.

## A real isolation bug the review caught

`R_ENVIRON=""` does **not** stop a user's `~/.Renviron` re-pointing the library — it gates only the *site* file; the per-user one is `R_ENVIRON_USER`. Verified with a throwaway `HOME`: before the fix, a `~/.Renviron` setting `R_LIBS_USER` won, and `.libPaths()[1]` was the user's own library despite activating.

`uvr run` was accidentally safe here because it passes `--no-environ`, which a sourced script cannot use. The fix lands in `REnv::vars()`, so both paths are now genuinely equivalent — which is exactly what sharing the builder (#177) was for.

Two other fixes from the same pass:

- **Python venv collision.** The unwind guard fired on `deactivate` merely existing. virtualenv defines a function by that name, so activating a uvr project inside a venv silently tore the venv down. Now gated on `UVR_PROJECT` in all three shells.
- **Child-process pollution.** `UVR_OLD_*` bookkeeping was exported, pushing a dozen-plus entries into every child including R. R now sees zero.

## Prompt (off by default)

```toml
[activate]
prompt = true
```

or `UVR_ACTIVATE_PROMPT=1` per-shell, which overrides the manifest either way — so you can opt out of a project that opts in. Off by default because a mutated prompt is contentious and many users build theirs from a framework that would fight with us.

## Testing

Verified live in **bash, fish, and PowerShell** — not just unit tests:

- project library lands first in `.libPaths()`, personal library correctly shadowed
- `deactivate` restores `PATH`, R library vars, and prompt exactly, then removes itself
- re-activation doesn't stack `PATH`; the PowerShell prompt prefix doesn't nest
- sourcing outside a project fails loudly, shell untouched
- a simulated Python venv survives activation

**474 tests pass**; `cargo fmt --check` and `clippy -D warnings` clean. New integration tests parse the emitted POSIX script with `sh -n`, so a syntax error can't ship.

One test deserves a note: `test_activate_powershell_isolation_survives_a_real_shell` runs wherever `pwsh` exists, **including the windows-latest runner**. Review raised that Win32's `SetEnvironmentVariable` deletes a variable assigned `''`, which would make the blanked isolation variables absent and leak the system site library. I confirmed PowerShell 7.6 on Linux does *not* do this, but I have no Windows to check the other side — so the property is pinned as a test that will fail in CI rather than in a user's session.

## Note on the pre-existing test failure

`sysreqs::tests::filter_missing_with_nonexistent_package` fails on my machine, but it also fails on pristine `b700a62` and this branch never touches `sysreqs.rs` — it's environmental (Arch, no dpkg/rpm/apk). Flagging rather than fixing, since it's outside this PR's scope.


gdevenyi:
Will return to the failing tests soon.

nbafrank:
Reviewed — the engineering held up under a hostile pass: the quoting is injection-safe in all three shells (single-quote literals, verified against paths with spaces/quotes/`$`), `set -u` clean throughout, the run.rs refactor into `REnv` is behavior-identical (diffed the env precedence line by line), and the zero-baked-paths shim design means activation can never go stale after an R bump. Nice work. Three things before merge:

1. **The red CI**: `test_activate_emit_every_shell_succeeds` and `test_activate_emit_sh_is_valid_shell` hard-require an ambient R, and the four CI jobs that run `cargo test --all` do so *before* the R-install smoke step — so they die at `find_r_binary` on every platform. Your own design doc (test plan section) said to `#[ignore]` the R-dependent round-trip, and your `have_pwsh()` gate two tests down is exactly the right pattern — apply it as a `have_r()` check (or bootstrap a portable R in-test). Note this also means the PowerShell isolation assertion has never actually run on Windows CI yet — it fails at R detection before ever invoking pwsh.
2. `deactivate`'s restore branch unconditionally `export`s (and `set -gx`es) restored variables. For `PS1` — which most users never export — that means an activate/deactivate cycle leaves `PS1` exported into every child process for the rest of the session. Restore export-state too, or special-case the prompt vars. The existing prompt test only checks the restored value, so it won't catch this.
3. Fish is the only emitter never handed to a real interpreter (`sh` gets `sh -n`, PowerShell gets a full round-trip) — a `fish -n`-gated syntax check analogous to the sh one would close that gap, and fish's emitter is the most structurally different of the three.

With those, this merges. Heads-up: main moved — #199 landed plus a bug-fix batch (52fcf6a); should rebase cleanly, nothing touches activate's files.

bsirak:
Hello @gdevenyi and @nbafrank. This is good work and it's very much appreciated, so please don't take my comment the wrong way. I'll try to be as brief as possible, but I'm happy to expand if necessary.

I don't believe this is the right solution to F0. Fundamentally, python as a language allows for isolation **BEFORE** interpretation, which is why `venv` (a python standard library module, not a `uv` module) works so well. That's not the case with R, so we need a different solution.

@nbafrank, I'm happy to jump on a call to discuss details if you'd like, as I'm not sure I'll be able to clearly articulate all of my thoughts in a comment. It potentially involves a refinement of the [uv feature-parity design](https://github.com/nbafrank/uvr/blob/main/docs/design/2026-07-23-uv-feature-parity-design.md), so let me know how you'd like me to start that conversation.

gdevenyi:
> @nbafrank, I'm happy to jump on a call to discuss details if you'd like, as I'm not sure I'll be able to clearly articulate all of my thoughts in a comment. It potentially involves a refinement of the [uv feature-parity design](https://github.com/nbafrank/uvr/blob/main/docs/design/2026-07-23-uv-feature-parity-design.md), so let me know how you'd like me to start that conversation.

Happy to see this as a discussion in an issue in the open.

bsirak:
> > @nbafrank, I'm happy to jump on a call to discuss details if you'd like, as I'm not sure I'll be able to clearly articulate all of my thoughts in a comment. It potentially involves a refinement of the [uv feature-parity design](https://github.com/nbafrank/uvr/blob/main/docs/design/2026-07-23-uv-feature-parity-design.md), so let me know how you'd like me to start that conversation.
> 
> Happy to see this as a discussion in an issue in the open.

As it should be. Apologies for any implication it wouldn't be.

gdevenyi:
Rebased onto main (`f97cca3`) — clean apart from the changelog anchor, nothing near activate's files, as you predicted.

All three points were addressed in `9bc5c33`; re-confirmed on the rebased branch rather than taking the commit message's word for it:

1. **Red CI.** `have_r()` now gates every R-dependent test — the emit round-trip, the sh syntax check, the fish check and the PowerShell isolation assertion — following the `have_pwsh()` pattern you pointed at. Your note that the PowerShell assertion had never actually run on Windows CI was right, and it's the reason the pwsh test now gates on both.
2. **`PS1` export leak.** `deactivate` no longer exports the restored prompt variable; `activate.rs` special-cases it on the way out, and a test asserts the emitted script contains no `export PS1`, which the value-only test couldn't catch.
3. **Fish.** `test_activate_emit_fish_is_valid_fish` hands the emitted script to a real `fish -n`, gated on `have_fish()`, closing the gap you identified as the most structurally different emitter.

`cargo test --all`, clippy and fmt are green on the rebase.

On @bsirak's point — that R can't isolate before interpretation the way Python does, so this may be the wrong shape for F0 — I agree it belongs in an open issue rather than buried in this PR's thread, and I'd rather that discussion resolve before this merges than after. Happy to rework or withdraw if it lands somewhere else.

bsirak:
@nbafrank but...why?

nbafrank:
Merged — thanks, this is a substantial piece of work and it held up under a hostile review pass.

Verified on the final head: the quoting is injection-safe in all three shells (single-quote literals, checked against paths with spaces, quotes and `$`), the emitters are `set -u` clean throughout, the `run.rs` refactor into `REnv` is behaviour-identical (I diffed the env precedence line by line), and the zero-baked-paths shim design means activation can't go stale after an R bump. All three review items are addressed and confirmed: the R-gated tests plus the new CI step that finally executes the PowerShell isolation assertion on Windows (it had never actually run), the `PS1` export leak with a regression test, and the real `fish -n` check.

Closes #177, #178, #179, #180 — F0 is done. Looking forward to the F1 plan.

bsirak:
@gdevenyi Nice work on this one, seriously. The quoting safety, the REnv unification, and finally exercising the PowerShell isolation assertion on real CI are exactly the kind of thing that's easy to get subtly wrong, and you clearly didn't.

@nbafrank, one thing I want to flag before it becomes the pattern F1 builds on, not as a complaint about what merged. Activation is scoped to whichever single shell process sources it. Every sibling (a second terminal tab, a Makefile target, a CI job, RStudio's own R session, an IDE that spawns R directly rather than through an activated shell) still resolves a bare `R`/`Rscript` independently and silently gets whatever's first on `PATH`, with no warning that it might differ from the "activated" one.

Digging into whether that's an acceptable tradeoff for the interactive-console case specifically, I don't think it is, because we don't actually need it there. `uvr run` with no script argument already drops into a fully isolated, correctly-versioned R console today, no sourcing required — it's a direct spawn, not shell state. So activation isn't closing a correctness gap for the person sitting at a terminal; the only thing left on its side of the ledger is not having to type `uvr run` before a bare `R`, plus the optional prompt prefix.

That's worth weighing against what it costs to keep: for #173's actual goal (`one R version, everywhere, reliably`) I'm already making the case for global & per-project trampolines/symlinks. This is the same approach `uv` uses for Python: a file at a fixed location that points straight at the right version; `uvr` rewrites that file whenever the version changes, so an IDE only has to be told where that fixed location is once, and it stays correct from then on, rather than a session you have to remember to enter. When that lands, it reaches IDEs, CI, and RStudio the way activation structurally can't, and gives you a bare `R` that's correct without sourcing anything. At that point, I don't see a functional gap activation is still covering, versus a cosmetic one.

I can write this up as an issue against #173, but I'm not sure I have an issue with the idea, just the implementation. So I'm not sure how to phrase it right now. If there's a case activation covers that `uvr run` plus a trampoline doesn't, something I'm not seeing, I'd genuinely like to hear it before I write that up, rather than waste everyone's time.

gdevenyi:
> one thing I want to flag before it becomes the pattern F1 builds on, not as a complaint about what merged. Activation is scoped to whichever single shell process sources it.

This is exactly my and all my user's workflows. Terminals, with possibly different projects, running different stacks with fixed versions. This is the same way uv thinks as far as I can tell. There are no system R version because I support scientists using HPC with bare environments using modules or something like UV

I don't quite understand what other use case you're thinking of, is it something like uv tool?

bsirak:
@gdevenyi, thank you for the insight, but now I think I'm more confused as to why this solution was chosen and merged. Here's why:

1. With regards to HPC clusters, aren't scheduler-submitted jobs (`sbatch job.sh`, PBS, etc.) run as non-interactive, usually non-login shells? The scheduler executes your script directly and it does not source `.bashrc/.profile` first. So there's no ambient moment where activation "just happens." You'd have to explicitly add source /path/to/project/.uvr/activate as its own line inside every job script, before the line that calls R/Rscript. That has to be true for every script. Compared to `uvr run analysis.R`, which replaces `Rscript analysis.R` as a single, self-contained, stateless call. There is no separate "activate first" line, nothing to forget, nothing that depends on shell type or whether the shell is interactive/login. It's a direct process spawn: resolve the right R, set the isolated env vars, and pass `--no-environ` as a hard process flag, all in one invocation with no session state involved at any point.  For HPC specifically, `uvr run` isn't just the safer option, it's the one that matches the execution model. Activation adds a step that has to be correctly present in every job template with no enforcement mechanism, for an environment that was never going to have a meaningful "activated session" to begin with.
2. `uv`'s core is built directly on official python specs, so there's no "thinking". PEP 440 (version identifiers), PEP 508 (dependency specifiers), PEP 517 (build-system-independent build frontend), and PEP 405 (virtual environments) are all implemented as standards-compliant Rust code. Project metadata lives in `pyproject.toml` following PEP 621, and its one-off script support is built on PEP 723 for inline dependency metadata. Its virtual environments are standards-compliant and work interchangeably with other tools so there's no lock-in or customization. This means the `.venv` (and by extension, it's `activate` script) uv creates is just a normal venv any other Python tool can use natively. `R` has no such equivalents.

It's slightly concerning that this PR, and #173 generally, purport to want `R-native uv feature parity` and the first feature to ship from it is not `R-native` and is does not bring `uv parity`. I know this isn't my repo and no one actually asked, but `uvr` should strive for `rig`, `renv` and `pak` parity all in one, like `uv` does for python, but not `uv` directly. It should do all of that without carrying any of the actual weight of `R`.

I'm sorry for the wall of text.

gdevenyi:
> With regards to HPC clusters, aren't scheduler-submitted jobs (sbatch job.sh, PBS, etc.) run as non-interactive, usually non-login shells? T

Yes, and they inherit the environment from the caller. We module load/source the runtime environment the job will run in and then submit the job.

I'm sorry but I'm not sure what all of this wall-of-text friction is about. R has genuinely terrible package and installation handling and python (via uv) has genuinely good handling.

My goal working on uvr is to make it "work alike" as much as possible as python.

PR #208, opened 2026-07-31
fix(sysreqs): read SystemRequirements from the downloaded tarball

### LINKED ISSUE #207: sysreqs: CRAN's PACKAGES has no SystemRequirements, so the Alpine local-rules fallback never resolves anything
`uvr sync --install-system-deps` on Alpine compiles CRAN packages from source and fails on missing system libraries, without printing any sysreqs warning.

## Repro

uvr 0.4.4, Alpine 3.24.1, R 4.6.

```sh
docker run --rm devxygmbh/r-alpine:4-3.24 sh -c '
curl -fsSL https://raw.githubusercontent.com/nbafrank/uvr/main/install.sh | UVR_INSTALL_DIR=/usr/local/bin sh
mkdir -p /tmp/tp && cd /tmp/tp
printf "Package: tp\nVersion: 0.0.1\nImports: terra\n" > DESCRIPTION
uvr lock
grep -c system_requirements uvr.lock
uvr sync --install-system-deps --library /tmp/lib
'
```

```
sysreqs in lockfile: 0
> Installing 2 package(s): 2 from source
  i  No binary repo for Alpine Linux 3.24.1 on R 4.6; compiling 2 package(s) from source.
 x ERROR  Failed to install terra
  configure: error: gdal-config not found or not executable.
```

`apk info -e gdal-dev` afterwards: not installed.
Same for `units` → `libudunits2.so was not found`.

## Cause

`system_requirements` is only ever populated from the resolved repo's `PACKAGES` index (`registry/cran.rs:650`), and CRAN's `src/contrib/PACKAGES` does not carry a `SystemRequirements:` field.
So every CRAN-sourced lockfile entry has `system_requirements: None`.

On Alpine that silently disables the entire #30 fallback:

1. PPM returns `Unsupported system` → `SysReqLookup::UnsupportedDistro` (`sysreqs.rs:120`).
2. `check_pkg_local()` early-returns on the `None` (`sysreqs.rs:286-288`), so no rule is ever matched.
3. `check.missing` stays empty and the "System dependency check skipped" warning is gated on `any_has_sysreqs` (`sync.rs:542-546`), which is false for the same reason — hence no output at all.

The vendored rules are fine and would resolve both cases: `gdal.json` → `gdal-dev`, `gdal-tools` and `udunits2.json` → `udunits-dev` on Alpine (unversioned constraints, so 3.24 matches), and the `apk add` path exists (`sync.rs:1567`).
The `SystemRequirements` string simply never reaches them.

This also explains the "uvr never warns about missing system deps" symptom reported for `xml2`/`libxml2-dev` in #30 — the rule table was not the problem.

## Where the field does exist

- The package tarball's own `DESCRIPTION`: `SystemRequirements: C++17, GDAL (>= 2.2.3), GEOS (>= 3.4.0), PROJ (>= 4.9.3), TBB, sqlite3`.
- P3M's source index `https://packagemanager.posit.co/cran/latest/src/contrib/PACKAGES` — which uvr already parses with the same code path — does carry `SystemRequirements:`.

Filling the field from P3M's source index when the resolved index omits it looks like the cheapest fix (one cached, conditional-GET index fetch, only on the fallback path); PR to follow.

Bioconductor's `PACKAGES` omits the field as well, so Bioc packages on unsupported distros have the same gap.

pat-s:
Fixes #207.

## Problem

`system_requirements` is only ever populated from the `PACKAGES` index of the repository a package resolved from (`registry/cran.rs:650`), and CRAN's `src/contrib/PACKAGES` carries no `SystemRequirements:` field.
Every CRAN-sourced package therefore reaches `check_pkg_local()` as `None`, where it early-returns before touching a single rule.

On a distribution Posit's sysreqs API doesn't cover this disables the whole #30 fallback: nothing is resolved, nothing is installed, and the "System dependency check skipped" warning stays silent too, because it is gated on some package declaring sysreqs (`sync.rs:542`) — which none do, for the same reason.
The visible result on Alpine is `uvr sync --install-system-deps` walking straight into a source build that dies on `gdal-config not found` / `libudunits2.so was not found`, with no sysreqs output at all.

## Fix

Read the field where it is actually published: the package's own DESCRIPTION, inside the tarball uvr has just downloaded.

- The sysreqs check moves from before the download phase to just after it, still before anything is compiled or extracted.
- `inspect_tarball` — which already reads `Built:` and `NeedsCompilation:` out of the downloaded tarball for install-time classification — now also returns `SystemRequirements`, via the DCF parser so wrapped values survive (terra's spans two lines; a line-prefix scan would truncate it at `PROJ (>=` and lose the proj/sqlite matches).
- The lockfile value still wins when a repository does publish the field; the tarball is consulted only as the fallback.

No extra network access and no second download: the tarballs are already on disk at that point, and reading a DESCRIPTION out of one is the same scan `inspect_tarball` was doing anyway.
Reading the tarball also makes the answer authoritative for the exact locked version, which an index- or API-level lookup by package name wouldn't be.

Two consequences of the move worth calling out explicitly:

- Packages served from the package cache have no tarball in this run, so they contribute no `SystemRequirements` beyond what the lockfile carries. They're installed without compiling, so a missing build-time header can't be what breaks them.
- If every package in a sync is a cache hit, `plans` is empty and the check doesn't run at all. Same reasoning: nothing is downloaded, nothing is built.

`Downloader::new` now takes a cloned client, since the check that follows still needs one.

## Verification

Alpine 3.24.1 / R 4.6, patched binary built for musl, project with `Imports: terra, units`:

```
> Installing 3 package(s): 3 from source
  i  No binary repo for Alpine Linux 3.24.1 on R 4.6; compiling 3 package(s) from source.

 ! WARN  Missing system dependencies for 2 package(s)
  units needs: udunits-dev
  terra needs: gdal-dev, gdal-tools, geos-dev, proj-dev, sqlite-dev
> Running: apk add gdal-dev gdal-tools geos-dev proj-dev sqlite-dev udunits-dev
v System dependencies installed.

v Installed 3 package(s) in 5m14s
```

`apk info -e` afterwards: `gdal-dev`, `geos-dev`, `proj-dev`, `sqlite-dev`, `udunits-dev` all present, and `terra`, `units`, `Rcpp` in the library.

Before the patch the same project printed no sysreqs line at all and stopped at `configure: error: gdal-config not found or not executable.`, with `gdal-dev` still absent afterwards.

`cargo test --workspace` green, `cargo clippy --workspace --all-targets` clean, `cargo fmt --check` clean.
Three tests added: the wrapped-value parse, the `detect_sysreqs_from_tarball` wrapper (including "declares none" staying `None` rather than `Some("")`), and one asserting the end of the chain — that the string read off a terra tarball resolves to `gdal-dev`/`geos-dev`/`proj-dev`/`sqlite-dev` on `alpine-3.24.1`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,16 @@ release page on GitHub. Issue numbers reference https://github.com/nbafrank/uvr/
 
 Pure tracking section — fixes and small features land here between tags.
 
+- System dependencies are resolved on distributions Posit's sysreqs API
+  doesn't cover (#207). `SystemRequirements` was only ever read from the
+  resolved repository's `PACKAGES` index, and CRAN's index doesn't publish
+  that field — so on Alpine every CRAN package reached the vendored local
+  rules with nothing to match, and `uvr sync --install-system-deps` went
+  straight into a source build that failed on the missing library (`terra`
+  → `gdal-config not found`, `units` → `libudunits2.so was not found`)
+  without printing a sysreqs warning either. The check now runs after the
+  tarballs are downloaded and reads `SystemRequirements` from each one's
+  DESCRIPTION, which is where the field is actually published.
 - System dependencies are resolved on RHEL and its rebuilds (#209). uvr asked
   both sysreqs catalogs under the raw `/etc/os-release` identity, which neither
   speaks: Posit's API answers `Unsupported system` for `rhel`/`8.10` but
--- a/crates/uvr-core/src/installer/binary_install.rs
+++ b/crates/uvr-core/src/installer/binary_install.rs
@@ -664,6 +664,12 @@ pub struct TarballMeta {
     /// True iff DESCRIPTION explicitly states `NeedsCompilation: no`.
     /// Absent means uvr can't prove the package is pure-R — treat conservatively as source.
     pub pure_r: bool,
+    /// Raw `SystemRequirements:` field, continuation lines joined.
+    ///
+    /// The tarball is the only place this is reliably published: CRAN's
+    /// `PACKAGES` index omits the field, so it never reaches the lockfile
+    /// and the sysreqs fallback has nothing to match (#207).
+    pub system_requirements: Option<String>,
 }
 
 /// Inspect a downloaded `.tar.gz` for both `Built:` and `NeedsCompilation` in
@@ -710,6 +716,12 @@ pub fn inspect_tarball(tarball_path: &Path, package_name: &str) -> Option<Tarbal
                 }
             }
         }
+        // `SystemRequirements` routinely wraps across lines (terra's spans
+        // two), so it needs the real DCF parser rather than the line-prefix
+        // scan above — a truncated value would silently lose rule matches.
+        meta.system_requirements = crate::dcf::parse_dcf_fields(&buf)
+            .remove("SystemRequirements")
+            .filter(|s| !s.trim().is_empty());
         return Some(meta);
     }
     None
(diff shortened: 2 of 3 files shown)


nbafrank:
Looks good!

PR #210, opened 2026-07-31
fix(sysreqs): ask the sysreqs catalogs under the names they use

### LINKED ISSUE #209: sysreqs: RHEL gets no system dependencies — uvr queries both catalogs under names they don't use (rhel/8.10 vs redhat/8)
On RHEL, `uvr sync` resolves no system dependencies from either source — not from Posit's API, and not from the vendored rules.

## Cause

uvr passes the raw `/etc/os-release` identity to both catalogs (`sysreqs.rs`, `detect_linux_distro` → `"{id}-{version_id}"`), and neither is keyed that way.

UBI 8 reports `ID="rhel"`, `VERSION_ID="8.10"`. Posit's API rejects that pair but answers for `redhat`/`8`:

```
?pkgname=xml2&distribution=rhel&release=8.10   -> {"code":14,"error":"Unsupported system","payload":null}
?pkgname=xml2&distribution=redhat&release=8    -> {"requirements":[{"name":"xml2","requirements":{"packages":["libxml2-devel"], ...}}]}
```

The local fallback then can't cover for it either: no vendored rule mentions `rhel` — the RHEL entries are written as `"distribution": "redhat"` with `"versions": ["8"]`, so both halves miss (`rhel` ≠ `redhat`, and `8.10` isn't `8`; `truncate_to_minor` only strips a third component).

Same shape for the rebuilds: `rocky` / `almalinux` vs the rules' `rockylinux`, and `sles` vs `sle`.

## Symptom

```
 ! WARN  System dependency check skipped on rhel-8.10
  Posit's sysreqs catalog doesn't cover this distribution, and the local fallback had no applicable rules.
```

The message is accurate about the outcome but misleading about the reason — Posit's catalog does cover RHEL, uvr is just asking under a name it doesn't publish.

This was invisible until #207 was fixed: with no `SystemRequirements` reaching the resolver, the local rules had nothing to match anyway, so the RHEL naming gap never surfaced.

## Fix

Normalize the `(ID, VERSION_ID)` pair onto the vocabulary both catalogs use before either is consulted. PR to follow.

Note that only the RHEL family is keyed by major release — Ubuntu is `22.04`, openSUSE `15.6`, SLE `12.3` in both catalogs — so the truncation has to be family-scoped.

pat-s:
Fixes #209.

## Problem

`detect_linux_distro()` hands the raw `/etc/os-release` identity to both sysreqs catalogs, and neither is keyed that way on RHEL.
UBI 8 reports `ID="rhel"`, `VERSION_ID="8.10"`; Posit's API answers only for `redhat`/`8`, and every vendored RHEL rule is written as `"distribution": "redhat"` with `"versions": ["8"]`.

So a RHEL host gets `Unsupported system` from the API, then falls through to a local rule table where neither half matches (`rhel` ≠ `redhat`, and `truncate_to_minor` only strips a third component so `8.10` never becomes `8`).
Both sources come back empty and the user sees:

```
 ! WARN  System dependency check skipped on rhel-8.10
  Posit's sysreqs catalog doesn't cover this distribution, and the local fallback had no applicable rules.
```

Verified against the live API:

```
?pkgname=xml2&distribution=rhel&release=8.10   -> {"code":14,"error":"Unsupported system","payload":null}
?pkgname=xml2&distribution=redhat&release=8    -> {"requirements":[{"name":"xml2","requirements":{"packages":["libxml2-devel"], ...}}]}
```

## Fix

Map the pair onto the vocabulary both catalogs are written in, once, before either is consulted:

- `rhel` → `redhat`
- `rocky` / `almalinux` → `rockylinux` (same repos, same package names as Rocky's entries)
- `sles` → `sle`, `opensuse-leap` / `opensuse-tumbleweed` → `opensuse`
- RHEL family only: version truncated to its major release

Everything else passes through untouched. The truncation is deliberately family-scoped — Ubuntu is keyed `22.04`, openSUSE `15.6` and SLE `12.3` in both catalogs, so a blanket major-truncation would break them.

Doing this in `detect_linux_distro()` fixes both paths at once: RHEL now gets the authoritative API answer rather than only a local-rules approximation, and the local fallback works too when the API is unreachable.

One visible side effect: warnings that name the distro now say `redhat-8` rather than `rhel-8.10`.

## Verification

`cargo test --workspace` green, `clippy --all-targets` clean, `fmt --check` clean.
Four tests added: the RHEL-family mapping, the non-RHEL pass-through (Ubuntu/openSUSE/Debian/Alpine keep their full version), and one end-of-chain check that the normalized pair resolves `libxml2` → `libxml2-devel` and `GDAL` → `gdal-devel` against the vendored rules.

Independent of #207, though that one is what made this visible: with no `SystemRequirements` reaching the resolver, the local rules had nothing to match anyway and the naming gap never surfaced.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,11 @@ Pure tracking section — fixes and small features land here between tags.
   API and nothing from the local fallback either. The `(ID, VERSION_ID)` pair
   is now mapped onto the catalogs' vocabulary — `rhel` → `redhat`,
   `rocky`/`almalinux` → `rockylinux`, `sles` → `sle`, openSUSE variants →
-  `opensuse` — with the RHEL family truncated to its major release.
+  `opensuse` — with the RHEL family truncated to its major release. Oracle
+  Linux (`ol`) joins them: it is a RHEL rebuild that neither catalog knew.
+  Coverage still varies behind the mapping — the API serves `rockylinux` 9/10
+  but not 8, and rejects `fedora` and `alpine` outright — so the local rules
+  remain the backstop for those.
 
 ## v0.4.4 (2026-07-31)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,16 @@ release page on GitHub. Issue numbers reference https://github.com/nbafrank/uvr/
 
 Pure tracking section — fixes and small features land here between tags.
 
+- System dependencies are resolved on RHEL and its rebuilds (#209). uvr asked
+  both sysreqs catalogs under the raw `/etc/os-release` identity, which neither
+  speaks: Posit's API answers `Unsupported system` for `rhel`/`8.10` but
+  answers normally for `redhat`/`8`, and every vendored rule is written as
+  `redhat` with `versions: ["8"]`. RHEL hosts therefore got nothing from the
+  API and nothing from the local fallback either. The `(ID, VERSION_ID)` pair
+  is now mapped onto the catalogs' vocabulary — `rhel` → `redhat`,
+  `rocky`/`almalinux` → `rockylinux`, `sles` → `sle`, openSUSE variants →
+  `opensuse` — with the RHEL family truncated to its major release.
+
 ## v0.4.4 (2026-07-31)
 
 The biggest release yet, driven by a wave of new users: GitLab-hosted
--- a/crates/uvr-core/src/sysreqs.rs
+++ b/crates/uvr-core/src/sysreqs.rs
@@ -13,8 +13,9 @@ pub struct SysReq {
     pub package: String,
 }
 
-/// Detect the Linux distribution from `/etc/os-release`.
-/// Returns `(id, version_id)` like `("ubuntu", "22.04")`.
+/// Detect the Linux distribution from `/etc/os-release`, normalized onto the
+/// vocabulary the sysreqs catalogs use. Returns `id-version` like
+/// `"ubuntu-22.04"` or `"redhat-8"`.
 pub fn detect_linux_distro() -> Option<String> {
     let os = crate::os_release::detect()?;
     // Both halves are required here, unlike P3M slug detection: the sysreqs
@@ -25,7 +26,45 @@ pub fn detect_linux_distro() -> Option<String> {
     if os.id.is_empty() || os.version_id.is_empty() {
         return None;
     }
-    Some(format!("{}-{}", os.id, os.version_id))
+    let (distribution, release) = normalize_distro(&os.id, &os.version_id);
+    Some(format!("{distribution}-{release}"))
+}
+
+/// Map an `/etc/os-release` `(ID, VERSION_ID)` pair onto the names Posit's
+/// sysreqs API and the vendored `r-system-requirements` rules are written in.
+///
+/// Both catalogs speak the Posit vocabulary, not the os-release one, and both
+/// reject the raw values RHEL reports (#209):
+///
+/// ```text
+/// ?distribution=rhel&release=8.10  -> {"code":14,"error":"Unsupported system"}
+/// ?distribution=redhat&release=8   -> {"requirements":[{"name":"xml2", ...}]}
+/// ```
+///
+/// The vendored rules agree — every RHEL entry is written as `redhat` with
+/// `versions: ["8"]`, so a `rhel` / `8.10` host matched no rule either and the
+/// local fallback had nothing to offer once the API had bowed out. RHEL hosts
+/// therefore got no system dependencies from either source.
+///
+/// Only the RHEL family is truncated to its major: Ubuntu is keyed `22.04`,
+/// openSUSE `15.6` and SLE `12.3` in both catalogs.
+pub(crate) fn normalize_distro(id: &str, version_id: &str) -> (String, String) {
+    let distribution = match id {
+        "rhel" => "redhat",
+        // The RHEL rebuilds share Rocky's entries — same repos (crb /
+        // powertools), same package names.
+        "rocky" | "almalinux" => "rockylinux",
+        "sles" => "sle",
+        "opensuse-leap" | "opensuse-tumbleweed" => "opensuse",
+        other => other,
+    };
+    let release = match distribution {
+        "redhat" | "rockylinux" | "centos" | "fedora" => {
+            version_id.split('.').next().unwrap_or(version_id)
+        }
+        _ => version_id,
+    };
+    (distribution.to_string(), release.to_string())
 }
 
 /// Response from the Posit Package Manager sysreqs API.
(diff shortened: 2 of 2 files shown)


gdevenyi (commented):
Fetched the branch, ran `cargo test -p uvr-core --lib sysreqs` (18 pass), and independently checked the claim against the live API and the vendored rules.

**The fix is correct.** The vocabulary claim holds up:

| query | result |
|---|---|
| `rhel`/8.10, `rocky`/9, `almalinux`/9, `ol`/8, `centos`/9 | `Unsupported system` |
| `redhat`/8, `redhat`/10, `rockylinux`/9, `opensuse`/15.6, `sle`/12.3, `sle`/15.6, `centos`/8 | answers |

The vendored rules agree — the only distributions that appear anywhere in `rules/` are `redhat`, `centos`, `rockylinux`, `ubuntu`, `debian`, `fedora`, `opensuse`, `sle`, `alpine`. The family-scoped truncation is the right call: `redhat`/`rockylinux`/`centos` key on majors, `ubuntu`/`opensuse`/`sle`/`alpine` on dotted versions. Doing it in `detect_linux_distro()` covers the API and local paths at once, and I don't see a regression for any distro that works today.

Four things worth addressing, none blocking:

**1. The same family table already exists in this repo.** `crates/uvr-core/src/r_version/downloader.rs:296-302` classifies the identical ID set for the P3M slug (`"rhel" | "rocky" | "almalinux"` → major truncation, `"opensuse-leap" | "sles"` → opensuse). The output vocabularies genuinely differ (`rhel-8` vs `redhat-8`, `opensuse-156` vs `opensuse-15.6`), so merging them probably costs more than it saves — but they're now two tables of the same domain knowledge with no cross-reference. A `// see also` comment in both directions would be enough; without it, the next person adding an ID fixes one path and silently leaves the other broken.

**2. Oracle Linux (`ol`) is the same bug, unfixed.** RHEL rebuild, `ID="ol"`, `VERSION_ID="8.10"`; the API rejects it and no vendored rule mentions it. One match arm (`"ol" => "redhat"`) — and it's missing from the downloader table too.

**3. CentOS Stream 9/10 still gets nothing.** `centos` is in the truncation list, but both catalogs stop at centos 8 — `?distribution=centos&release=9` returns `Unsupported system`. There's precedent in the codebase for the fix: `p3m.rs:402-403` already maps `centos-9`/`centos-10` to `rhel9`/`rhel10`. Mapping `centos` >= 9 to `redhat` would close it.

**4. `opensuse-tumbleweed` → `opensuse` is inert.** Tumbleweed's `VERSION_ID` is a datestamp (`20260731`), which matches nothing in either catalog. Harmless, but the arm reads as coverage that doesn't exist — worth a comment, or dropping it.

Nits: `normalize_distro` could return `(&'a str, &'a str)` — both halves are borrowable (`&'static` or a subslice of `version_id`) — dropping two allocations and simplifying the assertions to `("redhat", "8")`. And the join-then-`split_once('-')` round trip through `resolve_system_deps`/`check_pkg_local` stays fragile for any unmapped hyphenated ID; this PR happens to fix the two that mattered.

gdevenyi:
Correction to point 3 above, plus some measured data.

**CentOS Stream is not as broken as I said.** I claimed Stream 9/10 "still gets nothing". Wrong: the API rejects `centos`/9 and `centos`/10, but the vendored rules carry version-less `centos` constraints, so the local fallback does resolve — `resolve_local("libxml2 (>= 2.6.3)", "centos", "9")` returns `["libxml2-devel"]`. Stream users get local-rule results with a degraded-check warning, not silence. Mapping `centos` >= 9 to `redhat` would upgrade them from local-only to authoritative API answers, which is still worth doing, but it is an improvement rather than a gap fix.

While building out distro CI I captured `/etc/os-release` from every image in the matrix and probed the API for each resulting pair. Three things that bear on this PR:

**`rockylinux` 8 is not served by the API** — 9 and 10 are:

```
?distribution=rockylinux&release=8   -> {"code":14,"error":"Unsupported system"}
?distribution=rockylinux&release=9   -> {"requirements":[...]}
?distribution=rockylinux&release=10  -> {"requirements":[...]}
```

So the `almalinux`/`rocky` → `rockylinux` mapping lands on the local rules for the 8 series and on the API for 9/10. Both work — the local rules cover `rockylinux` 8 — but it means Rocky/Alma 8 users get the degraded-check warning even after this PR. Worth a line in the mapping comment so it doesn't read as a regression later.

**Fedora and Alpine are local-rules-only.** The API rejects `fedora` at every release, including the ones `systems.json` lists, and rejects `alpine` too. The two catalogs genuinely disagree about coverage, so any test that treats `systems.json` as a proxy for what the API serves will be wrong in both directions.

**Oracle Linux is unmapped on both axes**, confirmed from the real image (`oraclelinux:8` reports `ID="ol"`, `VERSION_ID="8.10"`):

```
detect_posit_distro_slug -> "ol-8.10" -> ppm_linux_codename -> None      (no binaries)
normalize_distro         -> ("ol", "8.10") -> resolve_local -> []        (no sysreqs)
```

Same for `amzn` (Amazon Linux 2023). Oracle is a straight RHEL rebuild, so `"ol" => "redhat"` here plus `"ol"` in the `downloader.rs` RHEL arm would close it in two lines; Amazon Linux is a judgement call rather than an oversight.

Also, minor: `archlinux:latest` reports `VERSION_ID="20260726.0.562117"` (a rolling snapshot), so the "Arch publishes no version" reasoning in `detect_linux_distro`'s comment holds for a bare-metal install but not for the container — Arch reaches the API as `arch`/`20260726.0.562117` rather than being skipped early.

pat-s:
Thanks — the measured pairs are much more useful than the assumptions I was working from. Pushed f2a1546 taking three of the four points:

**Oracle Linux** — added `"ol" => "redhat"` with a test. Deliberately only the sysreqs half: the `downloader.rs` RHEL arm you mention selects *binaries*, which is a different risk profile (it decides what gets linked against the host), so I'd rather that be its own change than ride along in a sysreqs PR. Happy to do it as a follow-up if you'd rather have both at once.

**Rocky/Alma 8** — documented on `normalize_distro`, along with fedora/alpine being API-rejected at every release. Your framing is right: the mapping routes to a name a catalog knows, which is not the same as coverage, and the local rules stay the backstop. Better to have that written down than to have someone read the degraded-check warning on Rocky 8 later as a regression from this PR.

**Arch's snapshot `VERSION_ID`** — my comment was wrong and is corrected. Arch containers do reach the catalogs as `arch`/`20260726.0.562117` and simply match nothing, rather than being skipped by the empty-version guard.

**CentOS Stream** — left alone, with a test pinning the current behaviour and the reasoning. Since uvr never runs the rules' `pre_install` commands, mapping Stream >= 9 to `redhat` looks safe and would upgrade it from local-only to API answers — but it needs a version boundary (7/8 keep their own epel/powertools entries), which is an argument I'd rather have in its own PR than fold into this one.

nbafrank:
Merged! Looks good on my end
Thank you guys!

Issue #175, opened 2026-07-28
`ubuntu-2204` fallback for unknown distros installs broken jammy binaries on Arch (sf/xml2)


atsyplenkov:
When I set up a fresh uvr project on Arch Linux and add geospatial packages, the install succeeds but the packages fail to load, because uvr silently treats Arch as Ubuntu 22.04 and installs P3M's jammy binaries, which link against system-library SONAMEs that don't exist on a rolling-release distro.

Environment: Arch Linux (rolling, `ID=arch` in os-release), uvr 0.4.3 (installed via the AUR `uvr-bin` package), system R 4.6.1, proj 9.8.1, gdal 3.13.1, libxml2 2.15.3.

Repro:

```sh
mkdir test && cd test
uvr init
uvr add sf terra xml2
uvr run R -e 'library(sf)'
```

This gives:

```
Error: package or namespace load failed for 'sf' in dyn.load(file, DLLpath = DLLpath, ...):
 unable to load shared object '.../.uvr/library/sf/libs/sf.so':
  libproj.so.22: cannot open shared object file: No such file or directory
```

`library(xml2)` fails the same way with `libxml2.so.2`. Scanning every `.so` under `.uvr/library` with `ldd` shows exactly which jammy SONAMEs are missing on Arch:

```
sf/libs/sf.so:      libproj.so.22 => not found,  libgdal.so.30 => not found
xml2/libs/xml2.so:  libxml2.so.2 => not found
```

gdevenyi:
I had this issue with arch as well. I think we need a fallback to a manylinux build instead, but the R packages don't offer as comprehensive a set as python does. I can test this locally on my arch.

gdevenyi:
Reproduced this on Arch (CachyOS) and opened #199 with a fix. @atsyplenkov your diagnosis was exactly right, including the `ldd` scan — that's precisely where it goes wrong.

**Root cause.** `detect_posit_distro_slug` fell back to `ubuntu-2204` for any distro it didn't recognize, and that slug maps to PPM's `jammy` repo. Confirmed here:

```
uvr doctor  →  P3M binary packages  available (Linux jammy)
uvr add xml2 →  Installing 3 package(s): 3 binary   (2.7s)
ldd xml2.so  →  libxml2.so.2 => not found
```

[...]

Issue #203, opened 2026-07-30
uvr sync fails to extract RcppParallel binary tarball: "numeric field was not a number: when getting cksum for"


cfljam:
Thanks Francesco for this fantastic project!  It works just like I hoped and I can enjoy R config again...

One minor issue as follows:

---------------------------------------

## Environment

- `uvr` 0.4.3 (also reproduced on 0.4.2, before `uvr self-update`)
- RHEL 9.8 (Plow), x86_64, kernel 5.14.0-687.30.1.el9_8
- R pinned via `.r-version`: 4.4.0
- Package: `RcppParallel` 6.1.1, binary tarball from the P3M RHEL 9 binary repo

## What happened

```
$ uvr sync
> Installing N package(s): N binary
 x ERROR  Failed to install RcppParallel
  Failed to read tgz entry for 'RcppParallel': numeric field was not a number:  when getting cksum for
```

Fails identically with `uvr sync --ignore-cache` (forces a fresh download), so it isn't a corrupted/partial cached download.

## Root cause

The downloaded tarball itself is valid — `tar tzvf`, GNU tar 1.34, and Python's `tarfile` module all read it without error. It's GNU tar format (`tarfile.format == GNU_FORMAT`) and contains symlink entries (type `2`) for the bundled TBB shared libraries, e.g.:

```
RcppParallel/lib/libtbb.so -> ...
RcppParallel/lib/libtbbmalloc.so -> ...
RcppParallel/lib/libtbbmalloc_proxy.so -> ...
```

`uvr`'s tar-reading code (presumably the `tar` Rust crate) appears to mis-parse this GNU-format archive — the "numeric field was not a number ... when getting cksum" error is characteristic of the reader losing header alignment (e.g. mishandling a GNU long-name/long-link or symlink entry) and then trying to interpret a non-header block as a header. Since `tar`/`tarfile`/GNU tar all parse the same bytes fine, the archive is spec-compliant and the bug is in `uvr`'s extraction path, not the upstream binary.

## Reproduction

```
$ cd <pipeline with RcppParallel as a dependency, .r-version = 4.4.0>
$ uvr sync
# fails as above

$ uvr cache clean --package RcppParallel
$ uvr sync
# fails identically — rules out cache corruption

$ uvr sync --ignore-cache
# fails identically — rules out corrupted cached download
```

Cached tarball for reference:
[...]

PR #211, opened 2026-07-31
ci: run the full test suite natively in every supported distro

gdevenyi:
**Stack position: merge last.** Merge order is #213 → #215 → #216 → #211. All four are rebased on `bacdfe5` and #212/#214 have landed, so nothing borrows another branch's commits any more. Verified by merging all four in that order: the only collision in the stack is the `## Unreleased` changelog anchor between #215 and #216, where both sides append. `cargo test --all` 563 passed, clippy clean, fmt clean. #211 goes last because its matrix data and design doc describe the world *after* #213 lands — the findings table credits it by number. A doc that claims delivered what is not yet merged reads as authority the same way a stale gap does.

Follows #209 / #175. Design doc: `docs/design/2026-07-31-distro-conformance-ci-design.md`.

## Why

uvr maps `/etc/os-release` onto three *independent* third-party vocabularies, and they name the same host differently:

| Axis | Path | Vocabulary |
|---|---|---|
| R interpreter build | libc + arch | manylinux / musllinux |
| P3M binary packages | `detect_posit_distro_slug_from_os_release` → `ppm_linux_codename` | `rhel-8` → `centos8` |
| System dependencies | `detect_linux_distro` → API + vendored rules | `redhat`/`8` |

Axis 1 is distro-independent, which is why it has never broken this way. Axes 2 and 3 live in two separate match statements that no test compares. #175 was axis 2, #209 was axis 3, and neither was reachable from a test running on `ubuntu-latest` — the only thing that knows what a distro reports is that distro.

## What

29 distributions, matrix generated from `crates/uvr-core/tests/distro_matrix.json`. Each image gets the **binary that ships** — nothing is built inside the containers — and then runs `uvr --version`, `doctor`, `r install`, a binary package install + `library()` load, and a source build that needs system dependencies resolved.

Both Linux targets are built once — the gnu one through the same `ci/build-linux-gnu.sh` the release workflow uses (#212) — and handed to every image; each container picks between them with the same rule `install.sh` applies, so the selection is under test too. Building the artifact under test differently from the artifact that ships is the exact substitution this workflow exists to catch. Building in the container would hide the failures that matter most: a build made there links against the libc that is present by construction, so a glibc floor above what the distro ships can never show up.

`docker run` with the tree mounted rather than a `container:` job: `actions/checkout` needs a glibc the musl and minimal images don't have, so this is one shape that works for all 29. Package-manager dialects (apt/dnf/microdnf/zypper/apk/pacman/yum) live in one function in `ci/distro-suite.sh`, so adding a distro is a JSON entry, not a line of YAML.

## It fails today, on three supported distros

The published v0.4.4 `x86_64-unknown-linux-gnu` asset does not start on Ubuntu 20.04, Debian 11 or UBI 8:

```
/s/uvr: /lib64/libc.so.6: version `GLIBC_2.34' not found (required by /s/uvr)
```

All three are distros uvr otherwise supports — P3M publishes `focal`, `bullseye` and `centos8` binaries, and the sysreqs catalog covers all three. The **musl asset runs on all three**, so `install.sh` handing glibc hosts the gnu build is what strands them. Worth fixing separately (raise the gnu build's floor, or fall back to musl when the gnu binary won't exec); this PR is the thing that notices.

## The #209 assertion

The images carry libxml2's runtime library but not its headers, so

```sh
UVR_INSTALL_SYSREQS=1 uvr add xml2 --no-binary
```

compiles only if the os-release parse, the catalog naming, the package-manager probe and the installer dialect are all right for that distro.

Verified locally against `rockylinux:9`, same assertion, both sides of the fix:

```
pre-#210   exit=1   fatal error: libxml/tree.h: No such file or directory
post-#210  exit=0   ⚠ xml2 needs: libxml2-devel → dnf install -y libxml2-devel → sysreqs-smoke-ok
```

The published v0.4.4 binary is on the failing side of that line — it predates #210, which is now on main — so the stage reproduces #209 against a real released artifact.

Worth noting the failing run printed **no warning at all**, not even "System dependency check skipped". An assertion that greps for that message would have passed it clean. Only doing the install for real caught it.

## Fixtures are captured, never written

`os_release` values come from `ci/capture-os-release.sh`, and the workflow re-checks them against the image on every run. #209 was a reasoned-about fixture — RHEL reports `VERSION_ID=8.10`, not `8` — and a table written at a desk makes every offline test pass while production stays broken.

Capturing them turned up three things the code gets wrong today. They are recorded in the matrix rather than fixed here:

- **Oracle Linux mapped nowhere on either axis.** `oraclelinux:8` reports `ID="ol"` → slug `ol-8.10` → no P3M codename, and `("ol","8.10")` → no local rules. Fixed since: sysreqs in #209, binaries in #214, and this matrix now carries the RHEL slug, codename and sysreqs pair for it rather than nulls.
- **The two sysreqs sources disagree about coverage.** The API serves `rockylinux` 9 and 10 but not 8, and serves no `fedora` or `alpine` at any release, all of which the vendored rules cover. Tracked as separate `api` / `local` booleans.
- **`pick_sysreqs_installer` knows only apk/dnf/apt-get.** On openSUSE, SLES and Arch it falls through to `apt-get`, which isn't installed, so `--install-system-deps` can't work there. Those distros assert the diagnosis instead of the install.

Two smaller ones: `archlinux:latest` sets `VERSION_ID` to a rolling snapshot (a bare-metal Arch install sets none), so it is matched as `"*"` rather than pinned; and `curl` is a package *conflict* on RHEL images shipping `curl-minimal`, so tools are requested by command rather than by package name.

Dropping the compiler from the base image — closer to what a real user has — turned up two more. R's `utils` calls `system(which ...)` in `.onLoad` and fails to load without `which`, which the minimal RPM images don't ship; arguably `uvr doctor` should say so. And Alpine's `build-base` ships no `pkgconf`, which R's anticonf configure scripts need to find `libxml-2.0` — without it the source build fails for a second, unrelated reason and hides the first.

## Cost

| Trigger | Images | Wall clock |
|---|---|---|
| PR touching `sysreqs*.rs`, `p3m.rs`, `downloader.rs`, `os_release.rs`, the vendored rules or the matrix | 6 (`lane: "pr"` — one per package-manager dialect plus musl) | ~8 min including the shared build |
| nightly cron | 29 | ~15 min at `max-parallel: 12` |
| `workflow_dispatch` | any subset by key | — |

Free minutes on a public repo. `fail-fast: false`, since every distro is an independent question. Nightly failures open an issue rather than turning the tree red: upstream images drift on their own schedule (`rockylinux:9` was 9.3, is now 9.8) and a red main nobody caused is a red main everyone ignores.

## Alpine reproduces open issue #207

The musl path found two things the glibc images couldn't.

The binary stage cannot apply on musl at all — P3M publishes no musl repo and the portable manylinux fallback needs glibc, so `uvr add` correctly compiles from source, on an image that deliberately has no compiler yet. musl hosts now skip that stage; their coverage is the sysreqs path, which is what #30 was about.

With that out of the way, the sysreqs stage fails on Alpine, and not because of this suite: `libxml2-dev` is never installed and uvr prints **no warning at all**. That is #207 — CRAN's `PACKAGES` carries no `SystemRequirements`, and on Alpine the local rules are the only sysreqs source, so the fallback has nothing to match. (#209's report assumed #207 was already fixed. It isn't.)

**These jobs fail red, deliberately.** A matrix whose job is to tell you which distros are broken should say so in the only way CI says anything. The Alpine entries carry a note pointing at #207 so whoever sees the red job knows what they're looking at, and the failure blocks like any other until #207 lands.

Verified on `alpine:3.21` with a musl binary built from main: `install.sh`'s rule picks the musl asset, `uvr --version`, `doctor` and the R install all pass, and the sysreqs stage fails exactly as #207 describes.

## Relationship to the other three PRs

This matrix is what found the problems the other three fix, so its data had to be realigned before merge — a matrix that describes gaps as open after they are closed reads as authority while being wrong.

| Row | Then | Now |
|---|---|---|
| `oracle-8`, `oracle-9` | `gap`, null codename, `ol`/`8.10` | `rhel-8` → `centos8`, sysreqs `redhat`/`8` (#209 + #214) |
| `rocky-8`, `almalinux-8`, `centos-stream9/10` | `api: false` | unchanged — still the catalog fact — with a note that #214 asks under the RHEL name, and that the vendored rules deliberately keep the unaliased one |
| `rhel-8`, `ubuntu-2004`, `debian-11` | — | noted as expected to fail at `uvr r install`: portable R needs glibc 2.34. #212 is what lets the CLI start there at all |

Left deliberately failing: those three at `r install`, and the three Alpine entries at sysreqs (#207). Both are real, and both are the matrix doing its job.

## Not in this PR

The offline half of the design — `distro_conformance.rs` and `uvr doctor --json` — which catches everything offline-detectable in under a second with no containers. Happy to add it here or send it separately.

### DIFF
--- a/.github/workflows/distro-suite.yml
+++ b/.github/workflows/distro-suite.yml
@@ -1,0 +1,161 @@
+name: Distro suite
+
+# The full test suite, compiled and run natively inside every Linux
+# distribution uvr supports.
+#
+# Why this exists: uvr maps /etc/os-release onto three third-party
+# vocabularies (portable-R target, P3M binary repo, sysreqs catalog) and the
+# names differ per axis. #175 and #209 were both "we asked under a name that
+# side doesn't use", and neither was reachable from a test running on
+# ubuntu-latest. The only way to know what a distro reports is to ask it.
+#
+# The matrix is generated from crates/uvr-core/tests/distro_matrix.json —
+# adding a distro means adding an entry there, not editing this file.
+
+on:
+  schedule:
+    # Nightly, after the catalogs' business hours. The full matrix is too slow
+    # to gate every PR and its failures are often upstream drift rather than
+    # anything a PR did.
+    - cron: "0 5 * * *"
+  workflow_dispatch:
+    inputs:
+      keys:
+        description: 'Space-separated distro keys, or "all" / "pr"'
+        default: "all"
+        type: string
+  pull_request:
+    # On PRs, only when something that could change distro identification moves.
+    paths:
+      - "crates/uvr-core/src/sysreqs.rs"
+      - "crates/uvr-core/src/sysreqs_rules.rs"
+      - "crates/uvr-core/src/os_release.rs"
+      - "crates/uvr-core/src/registry/p3m.rs"
+      - "crates/uvr-core/src/r_version/downloader.rs"
+      - "crates/uvr-core/vendor/r-system-requirements/**"
+      - "crates/uvr-core/tests/distro_matrix.json"
+      - "ci/distro-suite.sh"
+      - ".github/workflows/distro-suite.yml"
+
+concurrency:
+  group: distro-suite-${{ github.ref }}
+  cancel-in-progress: true
+
+env:
+  # One R version here, not the two the main CI matrix uses: this job's subject
+  # is the distro, and the R-version axis is already covered on ubuntu/macOS/
+  # Windows by ci.yml.
+  R_VERSIONS: "4.5.1"
+  SMOKE_PKG: xml2
+
+jobs:
+  # Which images to run: the `pr` subset on pull requests (one per package
+  # manager family plus musl), everything on a schedule, and whatever was asked
+  # for on a manual dispatch.
+  select:
+    name: Select distros
+    runs-on: ubuntu-latest
+    outputs:
+      entries: ${{ steps.pick.outputs.entries }}
+      count: ${{ steps.pick.outputs.count }}
+    steps:
+      - uses: actions/checkout@v4
+      - id: pick
+        run: |
+          set -euo pipefail
+          matrix=crates/uvr-core/tests/distro_matrix.json
+          want='${{ github.event_name == 'pull_request' && 'pr' || inputs.keys || 'all' }}'
+          case "$want" in
+            all) sel="$(jq -c '[.distros[]]' "$matrix")" ;;
+            pr)  sel="$(jq -c '[.distros[] | select(.lane == "pr")]' "$matrix")" ;;
+            *)   sel="$(jq -c --arg k "$want" '[.distros[] | select(.key as $x | ($k | split(" ")) | index($x))]' "$matrix")" ;;
+          esac
+          echo "entries=$sel" >> "$GITHUB_OUTPUT"
+          echo "count=$(jq 'length' <<<"$sel")" >> "$GITHUB_OUTPUT"
+          jq -r '.[].key' <<<"$sel" | tr '\n' ' '
+
+  suite:
+    name: ${{ matrix.entry.key }}
+    needs: select
+    if: needs.select.outputs.count != '0'
+    runs-on: ubuntu-latest
+    timeout-minutes: 45
+    strategy:
+      # Every distro is an independent question; one answer must not suppress
+      # the other 28.
+      fail-fast: false
+      max-parallel: 12
+      matrix:
+        entry: ${{ fromJSON(needs.select.outputs.entries) }}
+    steps:
+      - uses: actions/checkout@v4
+
+      # `docker run` rather than a `container:` job: actions/checkout runs a
+      # node binary that needs a glibc the musl and minimal images don't have.
+      # Checking out on the host and mounting the tree in gives one shape that
+      # works for all 29 images.
+      - name: Full suite in ${{ matrix.entry.image }}
+        run: |
+          docker run --rm \
+            -v "$PWD:/work" -w /work \
+            -e "R_VERSIONS=${R_VERSIONS}" \
+            -e "SMOKE_PKG=${SMOKE_PKG}" \
+            "${{ matrix.entry.image }}" \
+            sh ci/distro-suite.sh
+
+      # The identity the suite ran under must be the identity the matrix
+      # claims. Cheap, and it is the assertion that keeps the offline
+      # conformance tests honest — a fixture nobody re-checks is how #209
+      # survived.
+      - name: Identity matches the matrix entry
+        run: |
+          set -euo pipefail
+          docker run --rm --entrypoint "" "${{ matrix.entry.image }}" \
+            sh -c 'grep -E "^(ID|VERSION_ID)=" /etc/os-release' > got.env
+          got_id=$(sed -n 's/^ID=//p' got.env | tail -1 | tr -d '"')
+          got_ver=$(sed -n 's/^VERSION_ID=//p' got.env | tail -1 | tr -d '"')
+          want_id='${{ matrix.entry.os_release.id }}'
+          want_ver='${{ matrix.entry.os_release.version_id }}'
+          [ "$got_id" = "$want_id" ] || {
+            echo "::error::${{ matrix.entry.key }}: image reports ID=$got_id, matrix says $want_id"
+            exit 1
+          }
+          # "*" = rolling release, version deliberately not asserted.
+          if [ "$want_ver" != "*" ] && [ "$got_ver" != "$want_ver" ]; then
+            echo "::error::${{ matrix.entry.key }}: image reports VERSION_ID=$got_ver, matrix says $want_ver — recapture with ci/capture-os-release.sh"
+            exit 1
+          fi
+
+  # Upstream images drift on their own schedule — rockylinux:9 was 9.3 and is
+  # now 9.8. That is a maintenance ticket, not a broken build, so the nightly
+  # reports it instead of turning the tree red.
+  report-drift:
+    name: Report matrix drift
+    needs: [select, suite]
+    if: failure() && github.event_name == 'schedule'
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+    steps:
+      - uses: actions/checkout@v4
+      - name: Open or update the drift issue
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          RUN_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
+        run: |
+          set -euo pipefail
+          title="Distro suite failing on the nightly matrix"
+          body="The nightly distro suite failed: $RUN_URL
+
+          If the failure is an identity mismatch, an upstream image moved and the
+          matrix needs recapturing:
+
+              ci/capture-os-release.sh <key>
+
+          If it is a suite failure, that distro is genuinely broken for users."
+          existing=$(gh issue list --state open --search "$title in:title" --json number --jq '.[0].number // empty')
+          if [ -n "$existing" ]; then
+            gh issue comment "$existing" --body "$body"
+          else
+            gh issue create --title "$title" --body "$body"
+          fi
--- a/.github/workflows/distro-suite.yml
+++ b/.github/workflows/distro-suite.yml
@@ -119,10 +119,6 @@ jobs:
     if: needs.select.outputs.count != '0'
     runs-on: ubuntu-latest
     timeout-minutes: 45
-    # Entries carrying `known_failure` reproduce an open issue. They still run
-    # and still show what they found — they just don't block, until the issue
-    # they name is fixed and the field comes off.
-    continue-on-error: ${{ matrix.entry.known_failure != '' }}
     strategy:
       # Every distro is an independent question; one answer must not suppress
       # the other 28.
--- a/.github/workflows/distro-suite.yml
+++ b/.github/workflows/distro-suite.yml
@@ -119,6 +119,10 @@ jobs:
     if: needs.select.outputs.count != '0'
     runs-on: ubuntu-latest
     timeout-minutes: 45
+    # Entries carrying `known_failure` reproduce an open issue. They still run
+    # and still show what they found — they just don't block, until the issue
+    # they name is fixed and the field comes off.
+    continue-on-error: ${{ matrix.entry.known_failure != '' }}
     strategy:
       # Every distro is an independent question; one answer must not suppress
       # the other 28.
--- a/.github/workflows/distro-suite.yml
+++ b/.github/workflows/distro-suite.yml
@@ -49,7 +49,11 @@ env:
   # is the distro, and the R-version axis is already covered on ubuntu/macOS/
   # Windows by ci.yml.
   R_VERSIONS: "4.5.1"
-  SMOKE_PKG: xml2
+  # The binary stage tests repo *selection*, so it needs a package P3M builds
+  # for every repo; the sysreqs stage tests devel-package resolution, so it
+  # needs one with an external system library. See ci/distro-suite.sh.
+  BINARY_PKG: jsonlite
+  SYSREQS_PKG: xml2
 
 jobs:
   # Which images to run: the `pr` subset on pull requests (one per package
@@ -162,12 +166,20 @@ jobs:
         # there and let the earlier stages carry them.
         env:
           SKIP_STAGES: ${{ !matrix.entry.sysreqs.api && !matrix.entry.sysreqs.local && 'sysreqs' || '' }}
+          # Entries the matrix records as unsupported assert the refusal rather
+          # than skipping the stage — see `expect_fail` in ci/distro-suite.sh.
+          # Absent `expect` yields empty strings, which disables it.
+          EXPECT_FAIL_STAGE: ${{ matrix.entry.expect.stage }}
+          EXPECT_FAIL_MESSAGE: ${{ matrix.entry.expect.message }}
         run: |
           docker run --rm \
             -v "$PWD:/work" -w /work \
             -e "R_VERSIONS=${R_VERSIONS}" \
-            -e "SMOKE_PKG=${SMOKE_PKG}" \
+            -e "BINARY_PKG=${BINARY_PKG}" \
+            -e "SYSREQS_PKG=${SYSREQS_PKG}" \
             -e "SKIP_STAGES=${SKIP_STAGES}" \
+            -e "EXPECT_FAIL_STAGE=${EXPECT_FAIL_STAGE}" \
+            -e "EXPECT_FAIL_MESSAGE=${EXPECT_FAIL_MESSAGE}" \
             "${{ matrix.entry.image }}" \
             sh ci/distro-suite.sh
 
--- a/.github/workflows/distro-suite.yml
+++ b/.github/workflows/distro-suite.yml
@@ -99,11 +99,17 @@ jobs:
             target
           key: distro-suite-cargo-${{ hashFiles('**/Cargo.lock') }}
           restore-keys: distro-suite-cargo-
+      # The gnu build goes through the same script the release workflow uses,
+      # so the matrix exercises the artifact users actually receive — glibc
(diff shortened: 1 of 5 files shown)


gdevenyi:
Posit publishes the whole supported-distro list as one documented, versioned endpoint — `GET /__api__/status`, described in the OpenAPI spec at `/__api__/swagger/doc.json` (spec version tracks the server build). Pushed b0d37a7 to use it as this design's catalog oracle instead of the per-pair `xml2` probing originally planned.

Every Linux entry, live:

| name | distribution | release | binaryURL | sysReqs | binaries | hidden | arch |
|---|---|---|---|---|---|---|---|
| centos7 | centos | 7 | centos7 | ✓ | ✓ | | x86_64 |
| centos8 | centos | 8 | centos8 | ✓ | ✓ | hidden | x86_64 |
| rhel7 | redhat | 7 | centos7 | ✓ | ✓ | | x86_64 |
| rhel8 | redhat | 8 | centos8 | ✓ | ✓ | | x86_64 |
| rhel9 | rockylinux | 9 | rhel9 | ✓ | ✓ | | x86_64, arm64 |
| rhel10 | rockylinux | 10 | rhel10 | ✓ | ✓ | | x86_64, arm64 |
| sles12 / sles15x | sle | 12.3 / 15.x | opensuse4x/15x | ✓ | ✓ | mostly hidden | x86_64 |
| opensuse15…156 | opensuse | 15 … 15.6 | opensuse15x | ✓ | ✓ | 156 visible | x86_64 |
| xenial…resolute | ubuntu | 16.04 … 26.04 | codename | ✓ | ✓ | ≤20.04 hidden | noble/resolute + arm64 |
| buster | debian | 10 | buster | ✓ | **✗** | hidden | x86_64 |
| bullseye…trixie | debian | 11 … 13 | codename | ✓ | ✓ | 11 hidden | x86_64 |
| manylinux_2_28 | centos | 8 | manylinux_2_28 | ✗ | ✓ | | x86_64, arm64 |

Three things this settles:

**The 29 hand-set `sysreqs.api` flags in the matrix are all correct** — 29/29 agree with the endpoint, including the non-obvious ones (`rockylinux` 9 and 10 but not 8; no `fedora` or `alpine` at any release). They can now be derived rather than maintained.

**It confirms the SLES split** this repo already encodes: `sles156` is distribution `sle`, release `15.6`, `binaryURL: opensuse156` — its own sysreqs vocabulary, openSUSE's binaries. That's the judgement I'd argued a scraper would flatten, and the endpoint agrees with the hand-written mapping. So the forward direction stays hand-written; the *reverse* direction ("what does Posit support that uvr doesn't map?") should now read the endpoint instead of a hand-maintained ignore-list.

**It exposes an arm64 gap.** Only `rhel9`, `rhel10`, `noble`, `resolute` and `manylinux_2_28` carry arm64 builds, and P3M routes by the R User-Agent:

```
jammy          aarch64  ->  source   (no arm64 build)
noble          aarch64  ->  BINARY   bin/4.5-noble-arm64
rhel9          aarch64  ->  BINARY   bin/4.5-rhel9-arm64
manylinux_2_28 aarch64  ->  BINARY   bin/4.5-manylinux_2_28-arm64
```

`ppm_linux_codename` has no arch dimension, so on arm64 Ubuntu 22.04, Debian 12, RHEL 8 or openSUSE, uvr prefers a distro repo with no arm64 build and compiles everything — when the portable repo would have handed it binaries. Nothing breaks the way #175 did, since P3M degrades to source rather than serving the wrong architecture, but it is slow and invisible. Worth its own issue; an arm64 lane here is what would keep it honest.

gdevenyi:
Rebased onto main (`f97cca3`) and onto the updated #212, and the Alpine rows are no longer accurate — in the good direction.

They said the sysreqs stage fails and named #207. #208 landed, #207 is closed, so rather than edit the note to match the issue tracker I re-ran the suite on `alpine:3.21` against a binary built from current main:

```
>>> skipping the binary stage: no binary repo exists for musl
  xml2 needs: libxml2-dev
> Running: apk add libxml2-dev
sysreqs-smoke-ok
>>> distro suite complete                                   exit=0
```

So the distro this matrix was built to reproduce a bug on now goes green through the same assertion. Notes and the design doc's findings table updated to say that, with #208 credited.

The build job also picks up #212's checksum verification for the zig toolchain, since it calls the same `ci/build-linux-gnu.sh` the release does.

Remaining expected failures are unchanged and deliberate: `rhel-8`, `ubuntu-2004` and `debian-11` fail at `uvr r install` because the portable R builds need glibc 2.34 — #212 is what turns that from a linker error into a legible message.

gdevenyi:
Rebased onto main (`bacdfe5`) and addressed all four review comments. Three were right; one was a false positive that I hardened anyway. The suite failure was something else entirely.

**Every distro job was failing on a relative path, not on a distro.** `./dist/uvr-x86_64-unknown-linux-gnu/uvr: not found`, six images, and the notes I wrote last time about which failures were "expected" were reading a symptom. `DIST` defaults to `./dist` and every stage cds into a scratch project under `/tmp` before invoking `$UVR`, so the path stopped resolving the moment it did. `sh` reports that as exit 127 "not found", which looks exactly like a missing artifact — so the build job was the first place I looked, and it was fine. `DIST` is now resolved against `ROOT` once, at the top, while the script is still standing in it. Reproduced both directions before and after:

```
OLD:  uvr ok in /work
      ./dist/uvr-x86_64-unknown-linux-gnu/uvr: No such file or directory
NEW:  uvr ok in /work
      uvr ok in /tmp/scratch-x
```

**The sysreqs stage ran on entries the matrix records as having no sysreqs coverage.** Correct, and it is worse than a spurious failure — it asserts the opposite of what the entry says. `arch` and `amazonlinux-2023` carry `api: false, local: false` *on purpose*, and the stage asserts that uvr names this distro's libxml2 devel package, which for those two it correctly cannot. Skipped there through the `SKIP_STAGES` the suite already understands, so the earlier stages still carry them. Those two are the only entries with no coverage at all — `rocky-8`, `centos-stream9/10`, `fedora-*` and `alpine-*` are `api: false` but `local: true`, and still run the stage.

**`sysreqs_autoinstall_supported` listed `yum`, contradicting its own comment.** The comment above it already said "knows apk/dnf/apt-get"; the case list said otherwise. `pick_sysreqs_installer`'s last branch is an unconditional `apt-get`, so on a yum-only image uvr does not fall back to yum — it shells out to an `apt-get` that is not there. Those hosts belong on the diagnosis assertion with zypper and pacman.

**`capture-os-release.sh` iterated `.[]` over an object.** Right, and it does not degrade quietly the way the comment suggests — the matrix is `{_comment, distros}`, so jq errors on the first key:

```
$ jq -r '.[] | "\(.key)\t\(.image)"' crates/uvr-core/tests/distro_matrix.json
jq: error (at distro_matrix.json:589): Cannot index array with string "image"
```

Now `.distros[]`.

**The nested-quote report on `.yml:70` is a false positive.** `${{ }}` is substituted before the shell ever sees the line, so `want='${{ … }}'` reaches bash as `want='pr'` — the Select distros job has been passing on every run. But interpolating a dispatch input into a quoted shell line *is* a real injection surface, since `inputs.keys` is free text that can close the quote, so it now goes through `env:` instead. Same for the `SKIP_STAGES` addition.

Not fixed here, and worth its own issue: `query_r_version` is unusable against a *managed* Windows R — see the note on #216.

gdevenyi:
Suite results after the path fix: **6 failing → 2**, and one of the two is a real bug this workflow just earned its keep by finding.

```
alpine-321    pass  1m9s      rocky-9       pass  1m25s
arch          pass  24s       ubuntu-2204   pass  1m19s
opensuse-156  FAIL  1m25s     rhel-8        FAIL  21s
```

**`arch` confirms the sysreqs skip does the right thing** rather than just going quiet — it runs the binary stage and stops:

```
>>> package manager: pacman
>>> install.sh would fetch: uvr-x86_64-unknown-linux-gnu.tar.gz
binary-smoke-ok
>>> distro suite complete
```

**`rhel-8` is the documented expected failure**, now reading the way #212 intended:

```
  Distros below this floor — Ubuntu 20.04, RHEL 8, Debian 11 — are not supported by
  uvr's R installer. Use your system package manager's R, or build R from source.
```

Worth saying plainly though: `rhel-8` is in the `pr` lane, so as it stands every PR that touches these paths gets a permanently red check that means nothing, and a red check nobody reads is how a real failure hides. The matrix records the expectation in prose (`note`) but nothing acts on it. Either `SKIP_STAGES=r` for the three sub-2.34 entries, or an `expect: "r"` field the runner honours, would make the lane trustworthy. Not done here — it is a design call, not a fix.

**`opensuse-156` is new, and it is a genuine uvr bug.** It was never reachable before: the job died at the first `$UVR` call in the binary stage, so the stage never ran. Now it does:

```
> Installing 3 package(s): 3 binary
 ! WARN  Missing system dependencies for 1 package(s)
   xml2 needs: libxml2-devel
   -> Hint: Install with: apt-get install -y libxml2-devel
   Failed to install xml2
   Archive for 'xml2' is not a built binary package (no Meta/package.rds — a source
   tarball served as binary, or a truncated download).
```

Two separate problems in eight lines:

1. **uvr told a zypper host to run `apt-get`.** That is `pick_sysreqs_installer`'s unconditional final branch leaking into user-facing advice — the same fallback behind the `yum` fix above, except here it reaches the user as an instruction that cannot work. The suite's zypper/pacman path asserts uvr *names* the package, which it does, so the wrong installer name sails through.
2. **The repo selected for openSUSE 15.6 served a source tarball as a binary.** Not a mapping miss — the live catalog carries `opensuse156` with `binaries: true`, and the matrix expects exactly that codename — so uvr picked the right repo and still got source back. This is #175's shape: a "binary" install that quietly is not one, invisible on ubuntu-latest.

Neither belongs in a CI PR, so neither is fixed here. Both want their own issue.

nbafrank:
The design here is right and the detective work behind it is real — build the artifact the way the release does rather than `cargo build`, test the binary that actually ships, mount the tree for images that can't run checkout's Node, centralize the package-manager dialects, split nightly from PR lanes to control cost. I want this workflow.

**But the CI is currently reporting failures that aren't distro problems, and that's the thing to fix before it merges** — a conformance matrix whose job is signal quality can't cry wolf.

At the time I reviewed (run 30704835043) all six pr-lane jobs were red. Five failed identically:

```
ci/distro-suite.sh: line 217: ./dist/uvr-x86_64-unknown-linux-gnu/uvr: No such file or directory
```

Root cause is `ci/distro-suite.sh`: `DIST` (line 27) and therefore `UVR` (line 170) are relative paths, but the binary-package stage does `cd "$work"` (line 216) before invoking `"$UVR"`. You already capture `ROOT="$(pwd)"` at line 33 precisely because you knew the cwd moves — `$UVR` just never gets re-anchored to it. Everything before the first `cd` (`--version`, `doctor`, `r install`) passes; everything after resolves against the scratch dir. The musl lane fails the same way one stage later at line 248.

One-line fix: `DIST="$(cd "${DIST:-./dist}" && pwd)"` right after `ROOT="$(pwd)"`, or build `UVR` from `$ROOT`.

I see you're down to 2 failures now (opensuse-156 and rhel-8), so you may already be on this. The rhel-8 one is the genuinely interesting result: `glibc 2.28 is too old for portable R builds (need >= 2.34)` is a real platform limitation this matrix found that no unit test could have — exactly what it's for. Worth keeping visible as a documented known limitation rather than silencing.

Once the pr lane is actually green, this merges. No objection to the runner-minute budget — the path filtering and nightly split are sensible.

gdevenyi:
Both pr-lane failures fixed, and neither was a distro problem — same conclusion you reached about the first six, one layer down. All six `pr` entries now verified green **locally, with the shipped artifact in the real images**, not just reasoned about:

```
ubuntu-2204   exit=0   binary-smoke-ok  sysreqs-smoke-ok
rocky-9       exit=0   binary-smoke-ok  sysreqs-smoke-ok
arch          exit=0   binary-smoke-ok  (sysreqs skipped: no coverage, by matrix)
alpine-321    exit=0   sysreqs-smoke-ok (binary skipped: no musl repo)
opensuse-156  exit=0   binary-smoke-ok  was FAIL
rhel-8        exit=0   expected failure confirmed        was FAIL
```

The gnu binary is built by `ci/build-linux-gnu.sh` exactly as the build job does — `requires at most GLIBC_2.28` — because the v0.4.4 release binary is pre-#212 and dies in the loader on ubi8 before reaching anything interesting. Worth knowing if you reproduce this: the released artifact cannot be used to test the rhel-8 lane.

**rhel-8: assert the refusal, do not skip it.** You wanted the limitation kept visible; I wanted the lane trustworthy. `SKIP_STAGES=r` would have bought the second by giving up the first, so instead the three sub-2.34 entries carry an `expect` block and the stage asserts:

```json
"expect": { "stage": "r", "message": "is too old for portable R builds" }
```

```
> Installing R 4.5.1 for LinuxX86_64
>>> expected failure confirmed: is too old for portable R builds
>>> distro suite complete (no R on this platform, as recorded)
```

It turns red in **both** directions, which is the part that makes it worth more than prose: if uvr stops refusing (floor moved — update the matrix) and if it refuses for some other reason (#212's message regressing into something a user cannot act on). Then it stops, because nothing downstream runs without R. The message substring is version-agnostic on purpose — RHEL 8 reports glibc 2.28, Ubuntu 20.04 and Debian 11 report 2.31.

**opensuse-156 was the suite measuring Posit, not uvr.** I had this wrong in my last comment: I said the repo "served a source tarball as a binary" and implied openSUSE was special. It is not. P3M's Linux `PACKAGES` index lists every package *whether or not a binary was built*; the binary-vs-source choice happens at download time from the R User-Agent. So coverage is per-package-version and lags on the quieter repos:

```
xml2 1.6.0 (released 2026-06-22)   jammy=BINARY  opensuse156=source  bookworm=source
jsonlite 2.0.0                     jammy=BINARY  opensuse156=BINARY  bookworm=BINARY
```

`bookworm` is in that first row too — this was never an openSUSE story, it was an xml2-1.6.0 story. Gating the suite on it measured Posit's build queue.

The binary stage's subject is repo *selection*, so it now has its own package — `jsonlite`, which ci.yml already uses for exactly this and which P3M builds for every repo in the matrix. The sysreqs stage keeps `xml2`, where needing an external system library is the entire point. One knob became two because the two stages were always asking different questions.

**What that leaves open, deliberately.** The underlying uvr bug is real and worse than the CI symptom: uvr announces `Installing N package(s): N binary` from the index and then **hard-errors** when the download turns out to be source, instead of falling back to the source build it already knows how to do. Any host whose repo lags on a package version hits it — not openSUSE-specific, not arch-specific. `verify_built_package` is right to reject the archive (#203 is why it exists); the error path around it is what is wrong.

Alongside it, the zypper/apt-get bug, which is broader than I said last time — it is not only `pick_sysreqs_installer`. `uvr doctor` on that same openSUSE host prints:

```
- No C compiler found. Install with: sudo apt-get install build-essential
```

Both are recorded in the design doc's findings table. Neither is fixed here: a CI PR that fixes two installer bugs is not a CI PR, and you would be right to send it back.

Also fixed while testing: an `expect` block with no `message` would have matched *any* failure — the same class of bug this mechanism exists to prevent, so it now refuses at startup rather than passing a distro for the wrong reason.

Rebase is a no-op — already on `bacdfe5`. Merged all four PRs in order (#213 → #215 → #216 → #211): only collision in the stack is the `## Unreleased` changelog anchor between #215 and #216, where both sides append. `cargo test --all` 563 passed, clippy clean with `-D warnings`, fmt clean, shellcheck clean.

nbafrank:
Merged. Verifying all six pr-lane entries locally in the real images, with the shipped artifact, is the standard I'd hoped for — and it's the right answer to a workflow whose entire value is signal quality. Both failures being one layer down from the first six, and neither being a distro problem, is a good illustration of why this matrix has to be trustworthy before it can be useful.

Keeping the rhel-8 glibc result visible as an expected failure rather than silencing it is the right instinct too. That's a real platform limitation no unit test could have found, and it should stay legible.

This closes the gap that's been generating most of our incoming bug reports: platform defects were being found by users on their machines instead of by CI on ours. Thanks for building it.

PR #212, opened 2026-07-31
release: pin the Linux glibc floor at 2.28 instead of inheriting the runner's

gdevenyi:
**Stack position: merge third. Independent of #213/#214 — it touches only the release workflow — but #211 is rebased on this branch and calls `ci/build-linux-gnu.sh`, so it must land before #211.**

Fixes the glibc failures #211's distro matrix found.

## Problem

The published v0.4.4 `x86_64-unknown-linux-gnu` binary does not start on RHEL 8, Debian 11 or Ubuntu 20.04:

```
$ uvr --version
/usr/local/bin/uvr: /lib64/libc.so.6: version `GLIBC_2.34' not found (required by uvr)
```

All three are distros uvr otherwise supports — P3M publishes `focal`, `bullseye` and `centos8` binaries, and the sysreqs catalog covers all three. `install.sh` hands glibc hosts the gnu build, so those users get a binary that cannot run.

A binary inherits the glibc of whatever built it. The matrix already pinned the gnu targets to `ubuntu-22.04` to keep that floor down, and 22.04 still emits `GLIBC_2.34`. The obvious fix — move to an older runner — is gone: GitHub retired `ubuntu-20.04` and there is nothing below 22.04 left.

## Fix

`cargo zigbuild --target <triple>.2.28`. zig ships stubs for every glibc version, so the floor is pinned regardless of the host, and it cross-compiles aarch64 in the same step — which retires the `gcc-aarch64-linux-gnu` sysroot that had the identical problem on arm.

**2.28** matches `PPM_MANYLINUX_GLIBC_MIN` in `r_version/downloader.rs`, the glibc below which uvr can serve no binary R packages at all. No point running lower, no reason to exclude higher.

The build lives in `ci/build-linux-gnu.sh` rather than inline YAML so #211's matrix can build the artifact it tests the same way the release builds it — #211 now calls it, which is why it stacks on this PR.

## Two assertions, not two assumptions

Both regress silently, and neither is visible on the machine that built it:

- the maximum `GLIBC_` symbol version does not exceed the floor
- nothing outside the C runtime appears in `NEEDED`

The second caught a real hazard while writing this. Every C dependency here (lzma, bzip2, zstd, ring) vendors its source but prefers a system library when pkg-config finds one — so a builder with `libbz2-dev` installed emits a binary needing `libbz2.so.1.0`, a Debian SONAME RHEL does not ship. The artifact would then depend on which image built it. The build now takes the system out of that decision.

## Verification

The resulting binary runs on Ubuntu 20.04, Debian 11, UBI 8, Rocky 8, Ubuntu 24.04, Debian 13 and openSUSE Leap 15.5 — the first three being the ones that failed before.

```
>>> target/x86_64-unknown-linux-gnu/release/uvr requires at most GLIBC_2.28
>>> needs only: libm.so.6 libpthread.so.0 libc.so.6 libdl.so.2 ld-linux-x86-64.so.2
```

The aarch64 build is verified by symbol inspection only (max `GLIBC_2.28`, same NEEDED set); I have no arm64 emulation here to execute it.

## What this does not fix

`uvr r install` still cannot work on those distros. The portable R builds need glibc 2.34 and uvr says so explicitly:

```
Distros below this floor — Ubuntu 20.04, RHEL 8, Debian 11 — are not supported by
uvr's R installer. Use your system package manager's R, or build R from source.
```

That is upstream's floor, not something a build change can move. What changes is that a RHEL 8 user now *gets* that explanation — and can still use uvr against a system R, which `find_all`/`find_r_binary` already support — instead of a dynamic-linker error with no path forward.

So #211's matrix will still fail on RHEL 8, Debian 11 and Ubuntu 20.04 at the `r install` stage after this lands — its entries for those three carry a note saying exactly that. The failure is honest and worth keeping visible; deciding what those distros should be promised is a separate call.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -22,10 +22,16 @@ jobs:
             os: macos-latest
           - target: x86_64-apple-darwin
             os: macos-latest
+          # Built through ci/build-linux-gnu.sh, which pins the glibc floor
+          # instead of inheriting the runner's. The runner version no longer
+          # matters, which is the point — ubuntu-22.04 was chosen to keep the
+          # floor down and still shipped a binary needing GLIBC_2.34.
           - target: x86_64-unknown-linux-gnu
-            os: ubuntu-22.04
+            os: ubuntu-latest
+            zigbuild: true
           - target: aarch64-unknown-linux-gnu
-            os: ubuntu-22.04
+            os: ubuntu-latest
+            zigbuild: true
           - target: x86_64-unknown-linux-musl
             os: ubuntu-latest
           - target: aarch64-unknown-linux-musl
@@ -41,13 +47,6 @@ jobs:
         uses: dtolnay/rust-toolchain@stable
         with:
           targets: ${{ matrix.target }}
-
-      - name: Install cross-compilation tools (Linux ARM64)
-        if: matrix.target == 'aarch64-unknown-linux-gnu'
-        run: |
-          sudo apt-get update
-          sudo apt-get install -y gcc-aarch64-linux-gnu
-          echo "CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_LINKER=aarch64-linux-gnu-gcc" >> $GITHUB_ENV
 
       - name: Install MUSL toolchain
         if: matrix.target == 'x86_64-unknown-linux-musl'
@@ -68,7 +67,13 @@ jobs:
           restore-keys: release-${{ matrix.target }}-
 
       - name: Build release binary
-        run: ${{ matrix.use_cross && 'cross' || 'cargo' }} build --release --target ${{ matrix.target }} -p uvr
+        run: |
+          if [ "${{ matrix.zigbuild }}" = "true" ]; then
+            sh ci/build-linux-gnu.sh "${{ matrix.target }}"
+          else
+            ${{ matrix.use_cross && 'cross' || 'cargo' }} build --release --target ${{ matrix.target }} -p uvr
+          fi
+        shell: bash
 
       # Smoke test the freshly-built binary on the runners where we can —
       # i.e. the target architecture matches the host. Catches regressions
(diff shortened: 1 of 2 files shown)


gdevenyi:
Rebased onto main (`f97cca3`) and addressed the review comment — it was the right thing to flag, since this toolchain links the binaries users install, and a compiler fetched over the network and executed unverified is a straight path from a CDN compromise into a shipped artifact.

The tarball is now checked against a pinned SHA-256 from `ziglang.org/download/index.json` (both x86_64 and aarch64), **before** it is unpacked rather than after — `tar` runs code paths on the archive either way, and an unpacked tree is already a foothold. On mismatch the download is deleted and the build aborts.

Verified both directions, since a guard that never fires is indistinguishable from no guard:

```
$ ci/build-linux-gnu.sh x86_64-unknown-linux-gnu          # cold cache, real checksum
>>> fetching zig 0.13.0
>>> target/x86_64-unknown-linux-gnu/release/uvr requires at most GLIBC_2.28
>>> needs only: libm.so.6 libpthread.so.0 libc.so.6 libdl.so.2 ld-linux-x86-64.so.2

$ # same script with the pin corrupted to zeros
!!! zig 0.13.0 checksum mismatch
    expected 0000000000000000000000000000000000000000000000000000000000000000
    got      d45312e61ebcc48032b77bc4cf7fd6915c11fa16e4aad116b66c9468211230ea
exit=1                                                    # and no unpacked tree left behind
```

I stayed with a pinned hash rather than a marketplace action: this script is also called by #211's matrix build and is meant to run identically on a developer's machine, where an action isn't available.

nbafrank:
Merged, and this is what a good infrastructure PR looks like — it doesn't just claim a property, it asserts it in the build. The post-build `objdump` checks on both the max `GLIBC_*` symbol and the `NEEDED` set are the part I'd have asked for if they weren't already there: the "builder happened to have libbz2-dev installed" class of leak is exactly what silently produces binaries that die on the target distro.

Verified `PPM_MANYLINUX_GLIBC_MIN` matches the floor the script pins, and confirmed this doesn't touch the `manylinux_2_28` claim from #199 — different axis (host glibc for R *package* binaries vs. the glibc the uvr binary itself links against).

Issue #54, opened 2026-04-27
Provide an option to choose which R binary to download if autodetection fails


joelostblom:
When I try to install R on my system, it seems like `uvr` cannot detect that I'm using a Ubuntu derivative (PopOS):

<img width="958" height="277" alt="Image" src="https://github.com/user-attachments/assets/ef4a6ec9-73a6-4fbc-84c8-d4f404e24a07" />

However, if I go to https://docs.posit.co/resources/install-r.html#optional-install-system-dependencies-of-common-r-packages and follow the steps for the Ubuntu install, everything works as expected. I think it would be great it there was an option to select which distribution/arch to download if none can be find, either as an interactive prompt or command flag.

An alternative solution could be to improve autodetection of the Linux OS, but I suspect that this problem might be deeper than uvr since `rig` also fails (although is the installation rather than the resolution/downloading step that fails): 

<img width="958" height="132" alt="Image" src="https://github.com/user-attachments/assets/1beebe5f-bda6-4411-89eb-65d55d8f671f" />

There will likely always be some Linux variant that can't be autodetected though, so I think the option to manually specify is preferred.

nbafrank:
@joelostblom — fixed on \`main\` (commit c17e4dd). New \`--distribution <SLUG>\` flag on \`uvr r install\`:

\`\`\`sh
uvr r install 4.5.3 --distribution ubuntu-2204
\`\`\`

Valid slugs match Posit's CDN paths: \`ubuntu-2204\`, \`ubuntu-2404\`, \`debian-12\`, \`rhel-9\`, \`centos-7\`, \`opensuse-154\`, etc. (full list: https://packagemanager.posit.co/client/#/repos/cran/setup). The flag overrides whatever \`/etc/os-release\` detection produced. No-op on macOS / Windows since they use CRAN, not Posit.

PopOS in particular reports \`ID=pop\` so the auto-detect doesn't match — \`--distribution ubuntu-2204\` (or whatever Ubuntu base PopOS is tracking) gives you the matching .deb.

Will ship in the next batched release.

joelostblom:
That's great thank you! I think the `--distribution` flag will be helpful and since R installs don't happen that often it doesn't add much overhead. Since package installs happen more often, it would be nice to not have to specify the flag each time, so I offered some additional thoughts on that in #55 .

nbafrank:
[...]

PR #213, opened 2026-08-01
feat(p3m): pick the binary repo from Posit's live catalog, per architecture

gdevenyi:
**Stack position: merge first.** Merge order is #213 → #215 → #216 → #211. All four are rebased on `bacdfe5` and #212/#214 have landed, so nothing borrows another branch's commits any more. Verified by merging all four in that order: the only collision in the stack is the `## Unreleased` changelog anchor between #215 and #216, where both sides append. `cargo test --all` 563 passed, clippy clean, fmt clean. Nothing depends on it to compile; #215 was held behind it because they overlap in `downloader.rs`, and both of its review items are now closed.

## Problem

`ppm_linux_codename` maps a distro slug to a repo codename and has no notion of architecture — because nothing in uvr knew there was one to have. Posit's `/__api__/status` publishes it: only `rhel9`, `rhel10`, `noble`, `resolute` and `manylinux_2_28` carry arm64 builds.

P3M routes by the R User-Agent and degrades to **source** when a repo has no build for the caller's architecture, so this failed quietly:

```
jammy          aarch64  ->  source                                  (no arm64 build)
noble          aarch64  ->  BINARY  bin/4.5-noble-arm64
rhel9          aarch64  ->  BINARY  bin/4.5-rhel9-arm64
manylinux_2_28 aarch64  ->  BINARY  bin/4.5-manylinux_2_28-arm64
```

An arm64 host on Ubuntu 22.04, Debian 12, RHEL 8 or openSUSE was pointed at a repo with no arm64 build and compiled every package, while the portable repo — which does build for arm64 — sat unused. Nothing broke the way #175 did, since P3M serves source rather than the wrong architecture. It was just slow, and invisible.

## Fix

Read the catalog at runtime: one `GET /__api__/status`, cached daily beside the package indexes, and pick a repo that serves this distro on this architecture.

The lookup keys on the `(distribution, release)` pair `normalize_distro` already produces for sysreqs, and applies the same `catalog_alias` (#214). **Both Posit catalogs speak one vocabulary and have the same holes in it**, so uvr needs one translation from `/etc/os-release` rather than two tables to keep in sync — the endpoint lists `sles156` as distribution `sle`, release `15.6`, `binaryURL: opensuse156`, exactly the split #210 encoded by hand. Without the alias, Rocky/Alma 8 and CentOS Stream fall through to the static table and lose precisely the architecture awareness this PR adds.

Three cases, kept distinct on purpose:

| catalog says | uvr does |
|---|---|
| names a repo | use it |
| knows the pair, no build for this arch | authoritative no → portable repo, else source. The table gets **no second vote** — it cannot express "not on arm64" |
| never heard of the pair, or unreachable | compiled-in table, which knows aliases the catalog doesn't publish |

That middle row is the whole point, and it's why the fallback isn't simply "try the table next".

## The portable repo is excluded from distro lookups by name

Posit publishes `manylinux_2_28` as `distribution: "centos", release: "8"` — the same key real CentOS 8 has. A plain scan returns whichever the array lists first, so getting the right one today is luck rather than design; and on arm64, where `centos8` has no build, the portable entry is the only match and would be returned **without going through the glibc floor check that guards it**. The test pins this with the portable entry listed first in the fixture, so ordering cannot mask it.

`--distribution` still wins over the catalog: it's the user saying uvr's autodetection is wrong on this host, and the catalog lookup starts from that same autodetection (#54). Added `distro_override_is_set()` so the catalog path stands down.

## Tests

Unit tests cover the arch filter (`jammy`/arm64 → `None`, `rhel9`/arm64 → `rhel9`), `binaries: false` entries (Debian 10 is catalogued for sysreqs only), the portable-repo ambiguity above, and that a declining catalog is not overridden by the table.

`p3m_status_live.rs` (`#[ignore]`d, network) asserts the compiled-in table still agrees with the live catalog — it's what answers offline, and a stale offline answer is invisible until someone is offline. All 14 mapped slugs agree today. It also pins the aliased pairs from #214 against the platform catalog in both directions: if Posit starts publishing `rockylinux` 8 the alias becomes unnecessary, and if it stops publishing the RHEL names the alias points nowhere — either way CI says so.

`cargo test --all` green (350 unit + 9 integration binaries), clippy clean, fmt clean.

## Not in this PR

`uvr doctor` still reports from the static table; it's sync and this path is async. Worth following up, since doctor is where a user would look to see which repo they're getting.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -343,6 +343,11 @@ jobs:
       - uses: dtolnay/rust-toolchain@stable
       - name: Check every mapped repo still serves
         run: cargo test -p uvr-core --test p3m_repos_live -- --ignored --nocapture
+      # The catalog is what answers at runtime; the table is only the offline
+      # answer. A stale table is invisible until someone is offline, so the
+      # comparison has to run somewhere that is online on a schedule.
+      - name: Check the table still agrees with Posit's platform catalog
+        run: cargo test -p uvr-core --test p3m_status_live -- --ignored --nocapture
 
   audit:
     name: Security audit
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,8 +18,6 @@ Pure tracking section — fixes and small features land here between tags.
   tarballs are downloaded and reads `SystemRequirements` from each one's
   DESCRIPTION, which is where the field is actually published.
 
-
-
 - Binary package repos are chosen from Posit's live platform catalog, and the
   choice now accounts for the host architecture. uvr matched a hardcoded
   slug → repo table with no notion of arch, but only `rhel9`, `rhel10`,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,19 @@ Pure tracking section — fixes and small features land here between tags.
   tarballs are downloaded and reads `SystemRequirements` from each one's
   DESCRIPTION, which is where the field is actually published.
 
+
+
+- Binary package repos are chosen from Posit's live platform catalog, and the
+  choice now accounts for the host architecture. uvr matched a hardcoded
+  slug → repo table with no notion of arch, but only `rhel9`, `rhel10`,
+  `noble`, `resolute` and `manylinux_2_28` carry arm64 builds — and P3M
+  degrades to *source* when a repo has no build for the caller's architecture.
+  An arm64 host on Ubuntu 22.04, Debian 12, RHEL 8 or openSUSE was therefore
+  compiling every package while the portable `manylinux_2_28` repo, which does
+  build for arm64, sat unused. uvr now reads `GET /__api__/status` (cached
+  daily) and picks a repo that actually serves this distro on this
+  architecture. The compiled-in table still answers when the endpoint is
+  unreachable, and `--distribution` still wins over both.
 - Oracle Linux gets binary packages, not just system dependencies (#209
   follow-up). `ID="ol"` was aliased onto RHEL for sysreqs but not for the P3M
   slug, so OL hosts resolved their system libraries correctly and then
--- a/crates/uvr-core/src/r_version/downloader.rs
+++ b/crates/uvr-core/src/r_version/downloader.rs
@@ -229,6 +229,14 @@ pub fn set_posit_distro_override(slug: String) {
     let _ = DISTRO_OVERRIDE.set(slug);
 }
 
+/// True when the user pinned the distro with `--distribution`. The live
+/// platform catalog keys off `/etc/os-release`, which is precisely what an
+/// override says not to trust, so the catalog path stands down when this is
+/// set (#54).
+pub fn distro_override_is_set() -> bool {
+    DISTRO_OVERRIDE.get().is_some()
+}
+
 /// Detect the Posit CDN distro slug from `/etc/os-release`, or use the
 /// override set by [`set_posit_distro_override`] if any.
 ///
--- a/crates/uvr-core/src/registry/mod.rs
+++ b/crates/uvr-core/src/registry/mod.rs
@@ -4,6 +4,7 @@ pub mod forgejo;
 pub mod github;
 pub mod gitlab;
 pub mod p3m;
+pub mod p3m_status;
 
 use semver::Version;
 
--- a/crates/uvr-core/src/registry/p3m.rs
+++ b/crates/uvr-core/src/registry/p3m.rs
@@ -8,6 +8,7 @@ use tracing::{debug, warn};
 
 use crate::error::Result;
 use crate::r_version::downloader::Platform;
+use crate::registry::p3m_status;
 use crate::resolver::normalize_version;
 
 /// Pre-built binary package index from Posit Package Manager (P3M).
@@ -45,7 +46,11 @@ impl P3MBinaryIndex {
         bioc_release: Option<&str>,
         posit_distro_slug: Option<&str>,
     ) -> Self {
-        let Some(info) = platform_info(platform, posit_distro_slug, r_minor) else {
+        // Ask Posit which repo actually has builds for this distro *and* this
+        // architecture before falling back to the compiled-in table (#211).
+        let live_repo = live_linux_repo(client, platform).await;
+        let Some(info) = platform_info(platform, posit_distro_slug, r_minor, live_repo.as_ref())
+        else {
             // Unsupported platform combination: macOS x86 / Windows always
             // OK; Linux only when we recognise the distro. Falls through
             // to source — same behavior as before #55.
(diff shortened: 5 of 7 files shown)


gdevenyi:
Rebased onto main (`f97cca3`, so #208 is in) and addressed all three review comments.

**`is_known` required `binaries: true`.** Right catch, and the reasoning matters more than the outcome. Debian 10 is catalogued for sysreqs with `binaries: false`; treating that as "unknown" handed the question back to the static table, which is exactly the second-guessing of a catalog negative this module exists to prevent. The predicate is now "the catalog carries this pair at all" — `codename_for` still filters on `binaries`, so Debian 10 resolves to no distro repo and falls to the portable one, but now by the documented rule rather than by accident. Test flipped to assert `is_known(debian, 10)` **and** `codename_for(debian, 10) == None`, which pins both halves.

**`codename_for` could return the portable repo.** Already fixed in the pushed head — you reviewed the pre-amend commit. The fix is the one you describe: `manylinux_2_28` is excluded from distro lookups by name, and the test fixture lists it *before* the real `centos8` entry so array order can't mask a regression. Worth adding that on arm64 it wasn't order-dependent at all — `centos8` has no arm64 build, so the portable entry was the only match and would have been returned without passing the glibc floor check that guards it.

**"four entries" vs five.** Stale after I added the `centos8` row. Now says six and names what each one is there to exercise, so the next person editing the fixture knows which cases they'd be dropping.

gdevenyi:
Rebased onto main (`bacdfe5`) — clean, no conflicts. The three review comments were already addressed in the previous push and are unchanged: `is_known` is "the catalog carries this pair at all" with the portable repo excluded, `codename_for` filters `manylinux_2_28` out by name, and the fixture comment says six. `cargo test --all` green, clippy clean, fmt clean.

nbafrank:
Design approved — this is the right fix for the actual root cause of #175/#209 (two independent os-release-to-vocabulary tables drifting apart), and routing through the same `normalize_distro` pipeline as sysreqs means there's now one translation path instead of two.

I verified rather than assumed: curled `__api__/status` live (it serves what `p3m_status.rs` expects, and I checked that Posit's duplicate RHEL 9/10 entries — one under `rockylinux`, one literally named "rhel9 (unused alias)" under `redhat` — don't confuse `codename_for`'s first-match, since both resolve to the same `binaryURL`). The three-tier fallback ordering is correct and the two tests pin the directions that matter. `--distribution` correctly short-circuits before any network call. I also went looking specifically for a #175-class "picks the wrong repo and fails at `library()`" path and could not construct one.

**Two fixes before merge:**

1. **The safety net isn't wired to anything.** The pitch for `tests/p3m_status_live.rs` is that a stale offline answer is invisible until someone is offline — but `git diff main..HEAD -- .github/workflows/ci.yml` is empty, the test is `#[ignore]`d, and the existing `test-p3m-repos` job runs a different binary (`p3m_repos_live`). As shipped it only runs if a human remembers it exists. Add a CI step invoking `cargo test -p uvr-core --test p3m_status_live -- --ignored`, same pattern as the existing job.

2. **`is_known()` doesn't check `arch` or `binaries`, and every `Distro` field is `#[serde(default)]`.** Full-envelope breakage degrades correctly to the table. But a *partial* field rename upstream (say `arch` → something else) silently defaults it to `[]`, making `serves()` false for every entry while `is_known()` still reports the pair as known — and `live_linux_repo` treats known-but-not-serving as an authoritative no, sending a working distro to source builds. It never selects a *wrong* repo, so this isn't the #175 class, but it's a silent slow-path regression triggered by an upstream change unrelated to that host. Requiring `binaries` in the predicate (or treating an all-empty-`arch` response as unreachable) closes it.

Note #214 has landed on main, so the compile-time dependency is resolved — this should rebase cleanly.

gdevenyi:
Both items addressed in 52b1aaa (kept as a separate commit so the delta since your review is readable).

**The `#[serde(default)]` hole is real, and it is wider than the one field.** Your reasoning is what makes it worth fixing: a whole-envelope failure degrades correctly to the table, and this partial one does not. But `arch` is not the only field with that exposure — `os`, `binaries` and `binaryURL` are each `#[serde(default)]` too, and a rename of any of them makes `serves()` vacuously false for every entry while `is_known()` keeps reporting pairs as known. Same silent slow-path regression, same trigger, three more doors.

So the guard is on the whole parse rather than on one predicate:

```rust
fn usable(distros: &[Distro]) -> bool {
    distros.iter().any(|d| {
        d.os == "linux" && d.binaries && !d.arch.is_empty() && !d.binary_url.is_empty()
    })
}
```

A catalog in which *nothing* serves binaries anywhere is a parse that lost a field, not a Posit that stopped building — so it reads as unreachable and the static table answers, exactly as if the endpoint were down. It gates the cache-read path as well as the fetch, since a bad body cached by an older build would otherwise skip the check on every subsequent run.

**I did not add `binaries` to `is_known`**, deliberately, and I want to flag the conflict rather than silently pick a side: that is the predicate the previous review round had me *remove* it from. Debian 10 is catalogued with `binaries: false`, and requiring `binaries` there would put it back to "unknown", which hands the question to the static table — the second-guessing of a catalog negative this module exists to prevent. Your parenthetical alternative ("or treating an all-empty-`arch` response as unreachable") closes the hole without reopening that one, so that is the half I took.

The new test pins the failure by construction rather than by assertion-on-current-behaviour — it renames the field in the fixture and checks the three-way distinction that matters:

```rust
let renamed = SAMPLE.replace("\"arch\"", "\"architectures\"");
assert!(!distros.is_empty(), "it still parses — that is the problem");
assert!(is_known(&distros, "ubuntu", "22.04"), "still 'known'");
assert_eq!(codename_for(&distros, "ubuntu", "22.04", "x86_64"), None);
assert!(!usable(&distros), "so the catalog must be declared unusable");
```

**The safety net is wired now.** You are right that as shipped it only ran if someone remembered it existed. It joins the existing `test-p3m-repos` job rather than getting its own — that job is already "does the table still match Posit", already online, already on every PR, and a second `runs-on` for one `cargo test` is a runner-minute for nothing:

```yaml
      - name: Check the table still agrees with Posit's platform catalog
        run: cargo test -p uvr-core --test p3m_status_live -- --ignored --nocapture
```

Green against the live endpoint just now, including the arm64 plumbing guard:

```
test the_portable_repo_still_covers_both_architectures ... ok
distro repos with arm64 builds: ["rockylinux/9", "rockylinux/10", "ubuntu/24.04"]
test some_distro_repo_carries_arm64 ... ok
test the_static_table_matches_the_live_catalog ... ok
test the_aliased_pairs_resolve_in_the_platform_catalog_too ... ok
```

`cargo test --all` green, clippy clean, fmt clean. Rebase is a no-op — the branch is already on `bacdfe5`.

nbafrank:
Merged. Verifying an upstream catalog by hand is exactly the kind of thing that's easy to skip, so thanks for doing it properly — and for widening the `#[serde(default)]` guard rather than just patching the one field I pointed at. You're right that `os`, `binaries` and `binaryURL` have the same exposure and that guarding the whole parse is the correct shape; treating a catalog that serves nothing as unreachable is a better invariant than enumerating predicates.

The arch dimension is the part I hadn't appreciated. An arm64 host pointed at jammy silently compiling everything, when manylinux would have handed it binaries, is precisely the #175 failure mode in a different costume.

The live check now runs in CI, which was the other half of making this trustworthy.

PR #214, opened 2026-08-01
fix(sysreqs): close the last three RHEL-family alias gaps

gdevenyi:
**Stack position: merge first.** #213 depends on this (it calls `catalog_alias`) and will not compile without it. Nothing depends on the reverse.

Follow-up to #209, closing the last three RHEL-family gaps the `/__api__/status` sweep on #211 turned up. All for the same reason: the name uvr uses is not the name a catalog publishes.

## Oracle Linux got no binaries

#209 aliased `ID="ol"` onto RHEL for sysreqs but not for the P3M slug, so OL hosts resolved their system libraries correctly and then compiled every package — `ol-8.10` maps to no PPM codename. Half an alias is its own failure mode, and worse than none, because the working half makes it look handled.

Both axes now agree, and a test asserts they do rather than leaving the next person to notice.

## Rocky/Alma 8 and CentOS Stream 9/10 fell back for no reason

`/__api__/status` lists every platform Posit serves. The sysreqs API has `rockylinux` 9 and 10 but no 8, and `centos` 7 and 8 but no Stream 9 or 10:

```
?distribution=rockylinux&release=8  -> {"code":14,"error":"Unsupported system"}
?distribution=redhat&release=8      -> {"requirements":[{"name":"xml2", ...}]}
?distribution=centos&release=9      -> {"code":14,"error":"Unsupported system"}
?distribution=redhat&release=9      -> {"requirements":[{"name":"xml2", ...}]}
?distribution=centos&release=10     -> {"code":14,"error":"Unsupported system"}
?distribution=redhat&release=10     -> {"requirements":[{"name":"xml2", ...}]}
```

Those hosts were already getting correct answers — from the vendored rules, with a "check degraded" warning, for distros Posit knows perfectly well under their RHEL name.

## Why the alias lives at the query site, not in `normalize_distro`

Because the two vocabularies are **not** interchangeable in the local rules. Measured:

```
resolve_local("leptonica",  "rockylinux", "8") -> ["leptonica-devel"]
resolve_local("leptonica",  "redhat",     "8") -> []
resolve_local("libarchive", "centos",     "8") -> ["libarchive-devel"]
resolve_local("libarchive", "redhat",     "8") -> ["libarchive"]
```

Aliasing the local lookup as well would trade a spurious warning for wrong package names — `libarchive` instead of `libarchive-devel` is a source build that fails at `configure` with no hint why. So `catalog_alias` is applied where a catalog is queried, and `check_pkg_local` keeps the unaliased pair. A test pins all four of those resolutions so a future "simplification" that merges them fails loudly.

That measurement is also why this PR does **not** collapse the RHEL family onto `redhat` wholesale, which was the tidier-looking option: across 41 common `SystemRequirements` strings, `rockylinux` 9/10 and `redhat` 9/10 agree on everything but ordering, but `rockylinux` 8 differs from `redhat` 8 in 7 of 41 — in both directions.

The helper is named `catalog_alias`, not `api_alias`, and is `pub(crate)`: the platform list P3M binary selection reads has the same holes in the same vocabulary, so #213 applies it too. Keeping one copy is the point — two would drift.

## Tests

- `catalog_alias` maps the three unpublished pairs and leaves every published one alone, including `centos`/`stream`, which must not be mistaken for a release number.
- the local rules are asserted *not* aliased, via the four resolutions above.
- Oracle Linux resolves to `rhel-8` → `centos8` on the binary axis, alongside the existing sysreqs assertion.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,22 @@ Pure tracking section — fixes and small features land here between tags.
   without printing a sysreqs warning either. The check now runs after the
   tarballs are downloaded and reads `SystemRequirements` from each one's
   DESCRIPTION, which is where the field is actually published.
+
+- Oracle Linux gets binary packages, not just system dependencies (#209
+  follow-up). `ID="ol"` was aliased onto RHEL for sysreqs but not for the P3M
+  slug, so OL hosts resolved their system libraries correctly and then
+  compiled every package from source.
+
+- Rocky/Alma 8 and CentOS Stream 9/10 get authoritative sysreqs answers
+  instead of the local-rules fallback. Posit's catalogs serve `rockylinux`
+  from 9 on and `centos` only through 8, so those hosts fell back to the
+  vendored rules and printed a degraded-check warning for distros Posit does
+  cover, under their RHEL name. Catalogs are now asked under a name they
+  publish. The
+  local rules keep the unaliased name on purpose — they are not
+  interchangeable there: `rockylinux` 8 carries `leptonica-devel` where
+  `redhat` 8 carries nothing, and `centos` 8 wants `libarchive-devel` where
+  `redhat` 8 says `libarchive`.
 - System dependencies are resolved on RHEL and its rebuilds (#209). uvr asked
   both sysreqs catalogs under the raw `/etc/os-release` identity, which neither
   speaks: Posit's API answers `Unsupported system` for `rhel`/`8.10` but
--- a/crates/uvr-core/src/r_version/downloader.rs
+++ b/crates/uvr-core/src/r_version/downloader.rs
@@ -292,7 +292,12 @@ pub(crate) fn detect_posit_distro_slug_from_os_release(content: Option<&str>) ->
         }
         "debian" => format!("debian-{version_id}"),
         "centos" => format!("centos-{version_id}"),
-        "rhel" | "rocky" | "almalinux" => {
+        // Oracle Linux reports ID="ol" and is a straight RHEL rebuild, so it
+        // takes RHEL's binaries. Without this it fell through to the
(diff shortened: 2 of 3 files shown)


gdevenyi:
Rebased onto main (`f97cca3`) and addressed the review comment.

You're right that the test asserted only half its claim. It now pins both sides of the `libarchive` divergence:

```rust
assert!(resolve_local("libarchive", "centos", "8").contains("libarchive-devel"));
assert_eq!(resolve_local("libarchive", "redhat", "8"), vec!["libarchive"]);
```

That second line is the one that actually guards the design: a future "simplification" that aliases the local lookup as well would swap a headers package for a runtime one, and the resulting source build fails at `configure` with nothing pointing at the cause. Asserting only the `centos` side would have let that through.

nbafrank:
Merged. I re-verified every alias against the live API rather than taking the table on faith, and it matches exactly — including the cases you deliberately left unaliased:

```
rockylinux/8  -> Unsupported system   (aliased to redhat/8, which is 200)
centos/9, /10 -> Unsupported system   (aliased to redhat/9, /10, both 200)
rockylinux/9, /10 -> 200 directly     (correctly NOT aliased)
centos/8          -> 200 directly     (correctly NOT aliased)
```

The Oracle Linux dual-axis fix is the valuable part — #209 only did sysreqs, so `ol` was getting half an alias, which is worse than none.

Merge note: this needed a semantic conflict resolution against #220, which landed the same day. Your diff patched `resolve_system_deps`, the per-package function #220 replaced with a batched `fetch_sysreqs_index`; I re-applied the `catalog_alias` call inside the new function so the aliasing covers the batched request. Same intent, new call site — worth knowing if you rebase anything on top.

PR #215, opened 2026-08-01
feat(r): accept the rolling channels, and model per-platform availability

gdevenyi:
**Stack position: merge second, after #213.** Merge order is #213 → #215 → #216 → #211. All four are rebased on `bacdfe5` and #212/#214 have landed, so nothing borrows another branch's commits any more. Verified by merging all four in that order: the only collision in the stack is the `## Unreleased` changelog anchor between #215 and #216, where both sides append. `cargo test --all` 563 passed, clippy clean, fmt clean. Approved; held only for #213, which is now ready. The rebase onto it is a fast-forward.

Both gaps come from comparing uvr against what `cdn.posit.co` actually publishes.

## `devel` and `next` were rejected outright

They are published for **every** platform and architecture:

```
macos-arm64      devel 200      macos-arm64      next 200
macos-x86_64     devel 200      macos-x86_64     next 200
windows-x86_64   devel 200      windows-x86_64   next 200
manylinux-x86_64 devel 200      musllinux-arm64  next 200
```

Worth noting the r-builds enumeration script misses the macOS and Windows rows — its branches for those two platforms are guarded by `if v not in ("next", "devel")`, so those six URLs are never probed.

uvr filtered them out of `uvr r list --all` and `resolve_install_version` rejected them as invalid versions, so there was no way to install a development build at all. (I initially read the code as letting them through to a silent install — wrong: `is_plausible_r_version("devel")` fails before the floor check ever runs.)

They now install under their own name — `~/.uvr/r-versions/devel`, so a project pinning `devel` resolves to it — and are labelled everywhere they surface:

```
Available R versions
  - devel [unstable]
  - next [unstable]
  v 4.6.1 [installed]
  - 4.6.0
  ...
  -> Hint: `devel` and `next` are rebuilt continuously — installing one pins a moving target, not a release.
```

```
> Installing R devel for LinuxX86_64
! Warning: devel is a rolling channel, not a release: it is rebuilt continuously and
  reinstalling later gives you different R.
  -> Hint: Pin a numbered version for anything you need to reproduce.
```

They are deliberately kept **out** of the release list rather than sorted into it. `next` is not "newer than 4.6.1" — it's a different kind of thing, and letting it sort as newest would make it the answer to "latest available" in the not-found hint and a candidate for partial-version resolution (`uvr r install 4.5`). Hence `AvailableVersions { stable, rolling }` instead of one `Vec`.

## Windows lost R 3.6.3

Availability was one floor for two platforms, `MAC_WIN_MIN_R_VERSION = (4,1,0)` — right for macOS, wrong for Windows. HEAD-probing every version in `versions.json` on both:

```
3.0.0 … 3.6.2   windows=403  macos=403
3.6.3           windows=200  macos=403
4.0.0 … 4.0.5   windows=403  macos=403
4.1.0 …         windows=200  macos=200
```

Windows is not contiguous, so no `>=` bound describes it: `4.1.0` rejects a build that exists, `3.6.3` advertises thirty that don't. Replaced with a predicate used by **both** the listing filter and the install preflight, so they can't drift apart, plus an availability string so the error names what the platform has ("R 3.6.3, or R 4.1.0 or newer") rather than a floor it doesn't have.

## Tests

Per-platform availability at each boundary (3.6.2/3.6.3/4.0.5/4.1.0 on Windows, 3.6.3/4.0.5/4.1.0 on macOS, full range on Linux); channels are named-not-versioned, so `parse_r_version` and `is_real_r_version` both reject them — the invariant that keeps them out of version ordering.

Verified end to end on this machine: `uvr r install devel` downloads, extracts and registers, `uvr r list` shows it as `devel [uvr-managed]`, and `uvr r uninstall devel` removes it.

`cargo test --all` green, clippy clean, fmt clean — on this branch and on the five-way merge with #211–#214.

## Two bugs found while testing this, not fixed here

Both are independent of this PR and want their own change:

1. **`linux_is_musl()` false-positives on glibc hosts.** It returns true if `/lib/ld-musl-*` merely exists, which it does on any glibc box with the `musl` package installed. On this machine — CachyOS, glibc 2.44 — uvr downloaded the **musllinux** R build, and that R cannot run: `Error relocating /lib/libz.so.1: __snprintf_chk: symbol not found`. `install.sh` gets the same host right by asking `ldd --version`, so uvr's two libc detectors disagree with each other.

2. **A fresh install is never health-checked.** `download_and_install_r` verifies `bin/R` *exists* but never runs it; the `query_r_version` probe only guards the already-installed short-circuit path. So the broken install above reported success. `ci.yml` documents the opposite ("`r install` … validates `R --version` internally, so a clean exit means each portable build runs on this OS/arch") — that claim holds for a re-install, not a first install.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,21 @@ Pure tracking section — fixes and small features land here between tags.
   interchangeable there: `rockylinux` 8 carries `leptonica-devel` where
   `redhat` 8 carries nothing, and `centos` 8 wants `libarchive-devel` where
   `redhat` 8 says `libarchive`.
+
+- `uvr r install devel` and `uvr r install next` work, and say what they are.
+  The rolling channels are published for every platform and architecture, but
+  uvr rejected them as invalid versions and hid them from `uvr r list --all`.
+  They now install under their own name and are labelled `[unstable]` in the
+  listing and warned about at install time — they are rebuilt continuously, so
+  pinning one is not reproducible.
+
+- Windows can install R 3.6.3 again. Availability was modelled as a single
+  floor shared by macOS and Windows (`>= 4.1.0`), but the two platforms differ
+  and Windows is not contiguous: the CDN publishes 3.6.3, nothing from 3.6.4
+  through 4.0.5, then 4.1.0 onwards. A floor either rejected 3.6.3, which
+  exists, or advertised thirty versions that do not, so availability is now a
+  per-platform predicate.
+
 - System dependencies are resolved on RHEL and its rebuilds (#209). uvr asked
   both sysreqs catalogs under the raw `/etc/os-release` identity, which neither
   speaks: Posit's API answers `Unsupported system` for `rhel`/`8.10` but
--- a/crates/uvr-core/src/r_version/downloader.rs
+++ b/crates/uvr-core/src/r_version/downloader.rs
@@ -92,15 +92,55 @@ const PORTABLE_CDN: &str = "https://cdn.posit.co/r";
 /// Unified version index for the portable builds.
 const VERSIONS_JSON_URL: &str = "https://cdn.posit.co/r/versions.json";
 
-/// macOS and Windows portable builds start at R 4.1.0 — the CDN returns 403
-/// for earlier versions on both platforms (verified: R 4.0.5 → 403 on
-/// `macos` and `windows`, 200 on `manylinux_2_34`). Linux builds have no floor.
-const MAC_WIN_MIN_R_VERSION: (u32, u32, u32) = (4, 1, 0);
+/// Rolling channels the CDN publishes alongside numbered releases, on every
+/// platform and architecture.
+///
+/// They are not versions. `devel` and `next` are rebuilt continuously, so what
+/// a project pins today is not what it resolves to tomorrow — which is why
+/// they are accepted but always labelled, never silently treated as a release.
+pub const ROLLING_CHANNELS: &[&str] = &["devel", "next"];
 
-/// The minimum R version published on the portable CDN for `platform`, or
-/// `None` when the platform has no floor (Linux).
-fn portable_min_r_version(platform: Platform) -> Option<(u32, u32, u32)> {
-    (platform.is_macos() || platform.is_windows()).then_some(MAC_WIN_MIN_R_VERSION)
+/// True when `s` names a rolling channel rather than a release.
+pub fn is_rolling_channel(s: &str) -> bool {
+    ROLLING_CHANNELS.contains(&s)
+}
+
+/// Whether the portable CDN publishes `version` for `platform`.
+///
+/// Not a floor: Windows is not contiguous. The CDN has exactly one build below
+/// 4.1.0 — R 3.6.3 — and nothing else until 4.1.0, so a `>=` bound either
+/// excludes a build that exists (3.6.3) or advertises thirty that do not
+/// (3.0.0–3.6.2, 4.0.0–4.0.5). Verified by HEAD-probing every version in
+/// `versions.json` on both platforms:
+///
+/// ```text
+/// 3.0.0 … 3.6.2   windows=403  macos=403
+/// 3.6.3           windows=200  macos=403
+/// 4.0.0 … 4.0.5   windows=403  macos=403
+/// 4.1.0 …         windows=200  macos=200
+/// ```
+///
+/// Linux is published continuously from 3.0.0 on both architectures.
+fn portable_publishes(platform: Platform, version: (u32, u32, u32)) -> bool {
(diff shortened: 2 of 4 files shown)


gdevenyi:
Rebased onto main (`f97cca3`) and addressed all three comments.

**The warning described behaviour that doesn't happen.** Correct — an install that already exists short-circuits, so "reinstalling later gives you different R" was wrong twice over: it won't refetch, and what actually moves is what the *name* means, not what's on disk. Reworded to say this install is a snapshot of today's `devel`, with the uninstall-then-install line for moving it forward:

```
! Warning: devel is a rolling channel, not a release: it names whatever was built
  most recently, so this install is a snapshot of today's devel.
  -> Hint: Pin a numbered version for anything you need to reproduce. To move this
     one forward later: uvr r uninstall devel && uvr r install devel.
```

**The hint hard-coded the channel names** while the list above it was derived from the index — so if Posit ever dropped one, the hint would name a channel that isn't listed. Now built from `available.rolling`.

**`rolling` doc overstated.** It said "which every platform has", which is true of the CDN but not of the field, since it only carries what the index lists and can be empty. Reworded.

gdevenyi:
Rebased onto main (`bacdfe5`). One conflict, in the `## Unreleased` section of the changelog against the entries #209/#214 added — both sides append, so both are kept. The three review comments were already addressed in the previous push and survive the rebase unchanged: the rolling-channel warning describes the short-circuit accurately, the hint is built from `available.rolling`, and the field doc no longer claims every platform has one. `cargo test --all` green, clippy clean, fmt clean.

nbafrank:
Approved — I HEAD-probed `cdn.posit.co/r/` against every claim rather than trusting the table, and the per-platform predicate matches reality exactly, including the discontinuity that makes a naive `>=` floor wrong:

```
windows/R-3.6.3 -> 200    windows/R-3.6.2 -> 403    windows/R-4.0.5 -> 403    windows/R-4.1.0 -> 200
macos/R-4.0.5   -> 403    macos/R-4.1.0   -> 200    macos/R-3.6.3   -> 403
macos-arm64,windows /R-devel, /R-next -> 200 (published everywhere, as you say)
```

`uvr r install devel` works; `release` correctly isn't claimed (it 403s on the CDN and appears nowhere in the code). Existing pins and manifests are untouched — `resolve_install_version` keeps the same `pick_newest_matching` behaviour for concrete `X.Y[.Z]` strings, just against `.stable`.

Holding the merge only until #213 lands, since they overlap in `downloader.rs` and #213 has two open items. Should be a trivial rebase.

gdevenyi:
Nothing to change here — this is a status update so the queue position is on the record rather than in my head.

**#213's two open items are now closed** (52b1aaa), so the thing you were holding this behind is resolved: the live-catalog comparison runs in CI as a step on the existing `test-p3m-repos` job, and the `#[serde(default)]` hole is shut by a whole-parse `usable()` guard covering `os`, `binaries`, `arch` and `binaryURL` rather than just the one field — a rename of any of them has the same silent-slow-path shape.

**The rebase is as trivial as you predicted.** I merged the four PRs in order onto `bacdfe5` and checked rather than assumed:

```
✓ #213 feat/p3m-live-status
✓ #215 feat/r-version-channels          (downloader.rs auto-merges)
✗ #216 fix/libc-detection-...           CHANGELOG.md only — both sides append
✓ #211 ci/distro-suite
```

`downloader.rs` auto-merges in every ordering: #213 adds one hunk at `set_posit_distro_override`, this PR works in `fetch_available_versions` / `resolve_install_version` / the version tables, and #216 is up in `linux_is_musl` and down in `download_and_install_r`. No two of them touch the same lines. The only collision in the whole stack is the `## Unreleased` anchor between this and #216, where both sides append and keeping both is the resolution.

Merged all four with that one resolution: `cargo test --all` 563 passed / 0 failed, clippy clean with `-D warnings`, fmt clean.

This branch is already on `bacdfe5` so there is nothing to force-push; it will want one rebase after #213 lands, which is a fast-forward.

Thanks for HEAD-probing the CDN against the table rather than taking it — the 3.6.3 / 4.0.5 discontinuity is exactly the thing a `>=` floor gets wrong in a way no test on current versions would catch.

nbafrank:
Merged. The CDN probes in my earlier review matched your predicate exactly, including the Windows discontinuity — 3.6.3 published, then nothing until 4.1.0 — which is precisely the shape a single floor can't express without either rejecting a version that exists or advertising thirty that don't.

Labelling the rolling channels `[unstable]` and warning at install time is the right treatment: they're genuinely useful for testing against R-devel, and pinning one is genuinely not reproducible. Saying both is better than picking a side.

Only a CHANGELOG collision on merge, resolved by keeping both entries.

Issue #99, opened 2026-05-20
⚠ Warning: R 4.6.0 pinned via .r-version but not installed (installed: 4.6.0, 4.6.0). Run `uvr r install 4.6.0`.


mvuorre:
(Reported in #98 but lifting here as a separate issue.)

I am pinning a specific R version in a fresh uvr project but keep getting "⚠ Warning: R 4.6.0 pinned via .r-version but not installed (installed: 4.6.0, 4.6.0). Run `uvr r install 4.6.0`.":

```bash
~/Documents
❯ uvr init uvr
  · No .r-version pin — IDE config bound to system R at /usr/local/bin/R. Run `uvr r install <ver>` then `uvr r pin <ver>` for project-pinned R.
✓ Initialized project uvr
  uvr.toml
  .uvr/library/

~/Documents
❯ cd uvr

~/Documents/uvr
❯ uvr r install 4.6.0 && uvr r pin 4.6.0
› Installing R 4.6.0 for MacOsArm64
 INFO R 4.6.0 already installed at /Users/matti/.uvr/r-versions/4.6.0
✓ R 4.6.0 installed
  in 0ms
✓ Pinned R 4.6.0 → .r-version

~/Documents/uvr
❯ uvr sync
⚠ Warning: R 4.6.0 pinned via .r-version but not installed (installed: 4.6.0, 4.6.0). Run `uvr r install 4.6.0`.
 ✗ ERROR  No lockfile found. Run `uvr lock` to generate one.

❯ uvr lock
⚠ Warning: R 4.6.0 pinned via .r-version but not installed (installed: 4.6.0, 4.6.0). Run `uvr r install 4.6.0`.
✓ Lockfile updated — 0 package(s)
  resolved in 422ms

❯ uvr sync
⚠ Warning: R 4.6.0 pinned via .r-version but not installed (installed: 4.6.0, 4.6.0). Run `uvr r install 4.6.0`.
✓ Everything is up to date
  0 package(s) in 0ms
```

Here's `uvr doctor`:

```bash
❯ uvr doctor
⚠ Warning: R 4.6.0 pinned via .r-version but not installed (installed: 4.6.0, 4.6.0). Run `uvr r install 4.6.0`.
› uvr doctor

Platform
  ✓ OS / architecture            macos/aarch64
  ✓ P3M binary packages          available

R installations
  ✓ R 4.6.0                      /Users/matti/.uvr/r-versions/4.6.0/bin/R · managed
[...]

Issue #135, opened 2026-07-14
bug: TOCTOU race on R version dir under concurrent installs


gdevenyi:
**Severity:** HIGH
**Component:** R installer / concurrency
**Files:** `crates/uvr-core/src/r_version/downloader.rs:447-470`, `crates/uvr-core/src/r_version/manager.rs:49-57`

## Problem

Check-then-act pattern on the version directory with no locking:

```rust
if r_binary.exists() {
    if query_r_version(&r_binary).is_some() { return Ok(install_dir); }
    std::fs::remove_dir_all(&install_dir)?;
}
// ... later: std::fs::rename(staging, install_dir)
```

Two concurrent `uvr r install <same-ver>` processes both pass the check, both download, both race on the final rename. One loses; its staging dir is orphaned.

Same shape in `uninstall`: `exists()` then `remove_dir_all` — a concurrent install/uninstall can flip state between check and act.

Trigger is uncommon (typically only one install at a time) but parallel CI matrix jobs or `make -j` invocations can hit it.

## Suggested fix

Take an `flock` on a sibling lockfile (`.uvr-r-versions.lock`) for the duration of install/uninstall. Release on drop.

## Cross-references

- Audit context: comprehensive codebase review.

nbafrank:
Confirmed and fixed on main (bf35aa8): the final rename into ~/.uvr/r-versions/<ver> is now the arbitration point between concurrent installs — renaming onto a populated directory fails rather than clobbering, and the losing process adopts the winner's completed install instead of erroring (its staging dir is cleaned up). A concurrent uninstall of an already-removed version is likewise no longer an error. Ships in v0.4.4.

PR #216, opened 2026-08-01
fix(r): detect libc by asking ldd, and run R before calling it installed

gdevenyi:
**Stack position: merge third, after #215.** Merge order is #213 → #215 → #216 → #211. All four are rebased on `bacdfe5`. Verified by merging all four in that order: the only collision in the stack is the `## Unreleased` changelog anchor between #215 and this, where both sides append and keeping both is the resolution. `cargo test --all` 563 passed, clippy clean, fmt clean.

Two bugs found while testing #215. The second is why the first went unnoticed.

## A musl loader is not a musl host

`linux_is_musl()` returned true if any `ld-musl-*` existed under `/lib`. That says musl is **installed**, not that the host runs it — and installing Rust's `x86_64-unknown-linux-musl` target pulls musl in, so ordinary glibc developer machines claimed to be musl.

Observed on this machine (CachyOS, glibc 2.44, `musl` package present):

```
$ uvr r install devel
INFO Downloading R devel from https://cdn.posit.co/r/musllinux_1_2/R-devel-musllinux_1_2.tar.gz
v R devel installed

$ ~/.uvr/r-versions/devel/bin/R --version
Error relocating /lib/libz.so.1: __snprintf_chk: symbol not found
```

And the same predicate gates `ppm_manylinux_repo()`, which skips musl hosts — so these machines were silently getting **no binary packages either**, falling back to compiling everything.

Detection now asks `ldd --version` first. That's the same question `install.sh` asks when choosing which uvr binary to hand the host, and two detectors disagreeing about one machine is the shape of #175 and #209. The loader scan survives as a fallback for hosts without `ldd`, but glibc wins a tie.

The decision is a pure function over its four signals, so the regression is testable rather than dependent on whatever `/lib` the CI runner happens to have.

After, on the same host:

```
$ uvr r install 4.6.1
INFO Downloading R 4.6.1 from https://cdn.posit.co/r/manylinux_2_34/R-4.6.1-manylinux_2_34.tar.gz
$ ~/.uvr/r-versions/4.6.1/bin/R --version
R version 4.6.1 (2026-06-24) -- "Happy Hop"

$ uvr doctor
v P3M binary packages          available (Linux manylinux_2_28)
```

## A fresh install was never run

`download_and_install_r` checked that `bin/R` exists and stopped. The `R --version` probe guarded only the *already-installed* short-circuit — so the broken install above reported success, and `ci.yml`'s claim that "a clean exit means each portable build runs on this OS/arch" held for a re-install but not a first one. Which is exactly the case where a bad libc guess lands.

It now runs `R --version` on the new install, removes the directory when that produces nothing — otherwise the exists-only short-circuit adopts the broken tree on the next run and never reinstalls — and names the likely cause, since a bare relocation error from the loader tells a user nothing.

This is the check that would have caught the first bug on the machine where it happened, rather than three layers downstream.

## Why the check does not run on Windows

Worth stating in the description rather than only in the thread, because it is a
deliberate narrowing of the fix and not an oversight.

Adding this check turned `Test (windows-x86_64)` red: CI installed R 4.1.0, the
extract succeeded, the probe came back empty, and the check deleted the tree and
reported a failed install.

```
> Installing R 4.1.0 for WindowsX86_64
INFO Downloading R 4.1.0 from https://cdn.posit.co/r/windows/R-4.1.0-windows.zip
  The R build installed at C:\...\r-versions\4.1.0\bin\R.exe does not run on
  this host — starting it and asking for its version produced nothing.
```

That R runs fine, and #213/#215 install it on the same runner. What is new is
that this PR is the **first thing to ever probe a managed R on Windows**:
`find_all()` takes a managed install's version from the directory name
(`detector.rs:36-44`), never from the binary, so `query_r_version` had no Windows
coverage to lose.

Both obvious explanations are wrong. Pulling `bin/R.exe` out of the published zip
and reading it:

- **Not `R_HOME`.** `env_remove("R_HOME")` was the first suspect, but the
  front-end calls a `getRHOMEshort` at `0x4016b0` that derives R_HOME from
  `GetModuleFileNameA` and strips path components — it self-locates. The
  `getenv` → `HKLM\Software\R-core\R\InstallPath` variant at `0x401780` has no
  callers in `.text`; the registry path is dead code in this binary.
- **Not the dual-arch dispatch.** R 4.1.0 ships both `bin/i386/` and `bin/x64/`
  (4.5.1 is x64-only), so "picks i386 first" was a fair guess. It does not:
  `0x403195`, straight after the prologue, unconditionally writes `x64` into the
  arch buffer. `R_ARCH` and `--arch` overwrite it; nothing else does.

What is left is the hand-off itself. `bin/R.exe` is a front-end that rebuilds the
command line and re-spawns `bin\x64\R.exe` through msvcrt's `system()` —
`""%s\bin\%s\R.exe"` is in `.rdata` and `system` is in the import table — and our
probe passes `-e "cat(R.version$major, ".", R.version$minor, sep='')"`, which has
embedded quotes to survive that round trip. I could not take it further without a
Windows box, and I will not guess at a fix I cannot run.

**So the check runs on macOS and Linux only.** The failure it exists for is a libc
mismatch — this PR's whole subject, and a Linux question. Windows ships one build
per release and has no libc to mismatch, so gating the install path on a probe
that is demonstrably unreliable there trades a real bug for a hypothetical one.
macOS keeps the check; the probe is known-good there, which is how the #99 path at
`downloader.rs:598` was written in the first place.

**Deliberately not fixed here:** that same `:598` short-circuit probes with
`query_r_version` too, so on Windows an already-installed version fails the probe
today. That is a pre-existing bug this PR neither introduces nor repairs, and it
wants a Windows machine and its own change.

## Tests

- the regression directly: musl loader + glibc loader + `ldd (GNU libc)` → not musl; and with no `ldd` to ask, glibc still wins the tie.
- real musl hosts stay musl: Alpine by either marker, a non-Alpine musl distro via `ldd`, and a musl-only host with neither.
- `verify_r_runs` against two stand-in binaries: one that prints a version, one that exits 1 with a relocation error like the real failure. Asserts the message names libc.

`cargo test --all` green, clippy clean, fmt clean — on this branch and on the six-way merge with #211–#215.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,27 @@ Pure tracking section — fixes and small features land here between tags.
   interchangeable there: `rockylinux` 8 carries `leptonica-devel` where
   `redhat` 8 carries nothing, and `centos` 8 wants `libarchive-devel` where
   `redhat` 8 says `libarchive`.
+
+- Glibc hosts with musl installed get glibc R builds again. `linux_is_musl()`
+  answered yes whenever a musl dynamic loader existed anywhere under `/lib`,
+  which is true of any glibc machine that has the `musl` package — installing
+  Rust's `x86_64-unknown-linux-musl` target pulls it in, so this hit developer
+  boxes routinely. Those hosts were handed musllinux R, which dies in the
+  loader (`Error relocating /lib/libz.so.1: __snprintf_chk: symbol not found`),
+  and were also denied P3M binary packages entirely, since that path skips musl
+  hosts. Detection now asks `ldd --version` first — the same question
+  `install.sh` asks when it picks which uvr binary the host gets — and glibc
+  wins when both loaders are present.
+
+- A freshly installed R is checked before being called installed. The install
+  verified that `bin/R` exists but never ran it, so an R that could not start
+  was reported as a success and failed later somewhere less obvious. On macOS
+  and Linux `uvr r install` now starts the new R and asks it for its version,
+  then removes the install and says what likely went wrong if that produces
+  nothing. Windows is exempt: `bin/R.exe` there is a front-end that re-spawns
+  the real binary, and the probe returns nothing for a freshly unzipped
+  install that runs perfectly well.
+
 - System dependencies are resolved on RHEL and its rebuilds (#209). uvr asked
   both sysreqs catalogs under the raw `/etc/os-release` identity, which neither
   speaks: Posit's API answers `Unsupported system` for `rhel`/`8.10` but
--- a/crates/uvr-core/src/r_version/downloader.rs
+++ b/crates/uvr-core/src/r_version/downloader.rs
@@ -116,24 +116,100 @@ fn linux_portable_id() -> &'static str {
     }
 }
 
-/// True when the host uses musl libc — detected from `/etc/os-release`
-/// (`ID=alpine`) or the presence of a musl dynamic loader under `/lib`.
+/// True when the host runs musl libc.
+///
+/// Four signals, gathered here and weighed in [`musl_from_signals`]: Alpine's
+/// two self-identifications (`/etc/alpine-release`, `ID=alpine`), what `ldd
+/// --version` reports, and which dynamic loaders exist under `/lib` and
+/// `/lib64`.
 fn linux_is_musl() -> bool {
-    if let Ok(content) = std::fs::read_to_string("/etc/os-release") {
-        for line in content.lines() {
-            if let Some(val) = line.strip_prefix("ID=") {
-                if val.trim_matches('"').eq_ignore_ascii_case("alpine") {
-                    return true;
-                }
-            }
+    musl_from_signals(
+        alpine_marker_present(),
+        ldd_version_output().as_deref(),
+        loader_present("ld-linux-"),
+        loader_present("ld-musl-"),
+    )
+}
+
+/// Decide the host libc from independent signals, strongest first.
+///
+/// The signal that used to decide this alone — "a musl loader file exists" —
+/// says only that musl is *installed*, not that the host runs it. Any glibc
+/// machine with the `musl` package (a Rust `x86_64-unknown-linux-musl` target
+/// pulls it in, so this is common on developer boxes) claimed to be musl and
+/// was handed musllinux R builds, which then fail to start:
+///
+/// ```text
+/// Error relocating /lib/libz.so.1: __snprintf_chk: symbol not found
+/// ```
+///
+/// `ldd --version` is asked first because that is what `install.sh` asks when
+/// it chooses which uvr binary to hand the same host — two detectors that
+/// disagree about one machine is the shape of #175 and #209, and worth not
+/// repeating. The loader scan survives as a fallback for hosts without `ldd`,
+/// but glibc now wins a tie: a box with both loaders runs glibc.
+fn musl_from_signals(
+    alpine: bool,
+    ldd: Option<&str>,
+    glibc_loader: bool,
+    musl_loader: bool,
+) -> bool {
+    if alpine {
+        return true;
+    }
+    if let Some(out) = ldd {
+        let out = out.to_ascii_lowercase();
+        if out.contains("musl") {
+            return true;
+        }
+        if out.contains("gnu libc") || out.contains("glibc") {
+            return false;
         }
     }
-    std::fs::read_dir("/lib")
-        .map(|rd| {
-            rd.flatten()
-                .any(|e| e.file_name().to_string_lossy().starts_with("ld-musl-"))
+    if glibc_loader {
+        return false;
+    }
+    musl_loader
+}
+
+/// Alpine identifies itself twice; either is conclusive and neither costs a
+/// subprocess.
+fn alpine_marker_present() -> bool {
+    if std::path::Path::new("/etc/alpine-release").exists() {
+        return true;
+    }
+    std::fs::read_to_string("/etc/os-release")
+        .map(|content| {
+            content.lines().any(|line| {
+                line.strip_prefix("ID=")
+                    .is_some_and(|v| v.trim_matches('"').eq_ignore_ascii_case("alpine"))
+            })
         })
         .unwrap_or(false)
+}
+
+/// `ldd --version`, merging stderr: musl's ldd writes its banner there and
+/// exits non-zero, so both the status and the stream have to be ignored.
+fn ldd_version_output() -> Option<String> {
+    let out = std::process::Command::new("ldd")
+        .arg("--version")
+        .output()
+        .ok()?;
+    let mut text = String::from_utf8_lossy(&out.stdout).into_owned();
+    text.push_str(&String::from_utf8_lossy(&out.stderr));
+    (!text.trim().is_empty()).then_some(text)
+}
+
+/// Whether any dynamic loader whose name starts with `prefix` is installed.
+fn loader_present(prefix: &str) -> bool {
+    ["/lib", "/lib64"].iter().any(|dir| {
+        std::fs::read_dir(dir)
+            .map(|rd| {
+                rd.flatten()
+                    .any(|e| e.file_name().to_string_lossy().starts_with(prefix))
+            })
+            .unwrap_or(false)
+    })
 }
 
 /// Parse the host glibc version from `getconf GNU_LIBC_VERSION` (e.g. "glibc 2.39").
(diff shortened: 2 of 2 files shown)


gdevenyi:
Rebased onto main (`f97cca3`) and addressed both comments.

**The `linux_is_musl()` doc was stale** — it still described the two signals the old implementation used while the new one weighs four. Rewritten to name all of them and point at `musl_from_signals`, where the weighing actually happens.

**"`R --version`" was wrong throughout.** The probe is `query_r_version`, which runs `R --vanilla --slave -e "cat(R.version...)"`. That matters for exactly the reason you give: someone debugging a failed health check would run `R --version` by hand, quite possibly get output, and conclude the check is broken. Fixed in the function doc, the call-site comment, the error message and the changelog. I also fixed two pre-existing comments in the same file that said it — leaving those would have made the correction look arbitrary.

The error message now reads "starting it and asking for its version produced nothing", which is both accurate and what a user can act on.

gdevenyi:
Rebased onto main (`bacdfe5`), and the Windows job that was failing here was a genuine regression in this PR — the new health check deleted a working R.

**`verify_r_runs` failed on a perfectly good Windows install.** CI installed R 4.1.0 on windows-x86_64, the extract succeeded, and then the probe came back empty, so the check removed the tree and reported a failed install:

```
> Installing R 4.1.0 for WindowsX86_64
INFO Downloading R 4.1.0 from https://cdn.posit.co/r/windows/R-4.1.0-windows.zip
  The R build installed at C:\...\r-versions\4.1.0\bin\R.exe does not run on
  this host — starting it and asking for its version produced nothing.
```

That R runs fine; #213 and #215 install it on the same runner. What is new is that this PR is the **first thing to ever probe a managed R on Windows**: `find_all()` takes a managed install's version from the directory name (`detector.rs:36-44`), never from the binary, so `query_r_version` had no Windows coverage to lose.

I dug into why before deciding what to do, since the two obvious explanations both looked plausible and both are wrong. Pulling `bin/R.exe` out of the published zip and reading it:

- **It is not `R_HOME`.** `env_remove("R_HOME")` was my first suspect. But the front-end main calls a `getRHOMEshort` at `0x4016b0` that derives R_HOME from `GetModuleFileNameA` and strips path components — it self-locates. The `getenv`→`HKLM\Software\R-core\R\InstallPath` variant at `0x401780` has no callers in `.text`; the registry path is dead code in this binary.
- **It is not the dual-arch dispatch.** R 4.1.0 ships both `bin/i386/` and `bin/x64/` (4.5.1 is x64-only), so "picks i386 first" was a fair guess. It does not: `0x403195`, straight after the prologue, unconditionally writes `x64` into the arch buffer. `R_ARCH` and `--arch` overwrite it; nothing else does.

What is left is the hand-off itself — `bin/R.exe` is a front-end that rebuilds the command line and re-spawns `bin\x64\R.exe` through msvcrt's `system()` (`""%s\bin\%s\R.exe"` is right there in `.rdata`, and `system` is in the import table). Our probe passes `-e "cat(R.version$major, ".", R.version$minor, sep='')"`, which has embedded quotes to survive that round trip. I could not take it further without a Windows box, and I am not going to guess at a fix I cannot run.

**So the check now runs on macOS and Linux only.** The failure it exists for is a libc mismatch — that is this PR's whole subject, and it is a Linux question. Windows ships one build per release and has no libc to mismatch, so gating the install on a probe that is demonstrably unreliable there trades a real bug for a hypothetical one. macOS keeps the check; the probe is known-good there, which is how the #99 path at `downloader.rs:598` was written in the first place.

**One thing I deliberately did not fix here.** That same `:598` short-circuit probes with `query_r_version` too, so on Windows today an already-installed version fails the probe, gets deleted, and re-downloads 118 MB on *every* `uvr r install` — a pre-existing bug, same root cause, no relation to this PR. The tempting fix is to trust `exists()` on Windows, but that trades "always re-downloads" for "a genuinely broken install never self-heals", and picking between those on a root cause I could not finish pinning down does not belong in a PR about libc detection. Worth its own issue.

`cargo test --all` green (549 passed), clippy clean, fmt clean.

nbafrank:
The musl detection is solid — `ldd_version_output()` degrading to the loader-file fallback when `ldd` is absent is the right shape, glibc wins ties, and the no-`ldd` path is explicitly tested.

I also traced the interaction with the #135 rename race, since "verify it runs" plus "adopt the winner's tree" is exactly where a second race could hide: `verify_r_runs` runs after `install_r_portable` returns, unconditionally, on both the winner and the adopting-loser paths — so a loser can't inherit a false success from a winner whose build doesn't start. No new correctness bug; worst case is both processes spawning `R --version` and both attempting cleanup, which is wasted work rather than corruption.

**The open question is the Windows CI failure.** When I reviewed, `Test (windows-x86_64)` failed in the smoke step with your new check rejecting a fresh install:

```
The R build installed at C:\...\r-versions\4.1.0\bin\R.exe does not run on this host —
starting it and asking for its version produced nothing.
```

I checked the last five `main` runs and that same job installs the same `R-4.1.0-windows.zip` on the same runner image successfully every time — but main has no run-check, so it never probes. So it's either a real bug in `query_r_version`'s Windows invocation of a freshly-extracted `R.exe` (which your PR is the first thing to exercise), or R 4.1.0's Windows portable build genuinely can't start headless there and the old exists-only check was silently shipping a broken install. Both are worth knowing.

I see the checks are green now. If you root-caused it, could you say what it was in the PR? If it just stopped reproducing, I'd rather we understood why before merging a gate on the install path — a flaky verification step on every R install is worse than none.

gdevenyi:
It was root-caused, not outwaited — and you are right that the distinction is the whole question, so I have moved the analysis out of the thread and into the PR description where you asked for it.

Short version: **the check now runs on macOS and Linux only**, and that is the change that turned Windows green. Nothing about the probe became more reliable.

The failure was real and it was mine. This PR is the first thing in uvr to ever probe a *managed* R on Windows — `find_all()` takes a managed install's version from the directory name (`detector.rs:36-44`), never from the binary — so `query_r_version` had no Windows coverage to lose, and adding a gate on it deleted a working R 4.1.0.

I chased it far enough to rule out both explanations I would otherwise have shipped a guess against. Reading `bin/R.exe` out of the published zip:

- **Not `R_HOME`.** `env_remove("R_HOME")` was my first suspect. The front-end calls a `getRHOMEshort` at `0x4016b0` that derives R_HOME from `GetModuleFileNameA` and strips components — it self-locates. The `getenv` → `HKLM\Software\R-core\R\InstallPath` variant at `0x401780` has no callers in `.text`.
- **Not the dual-arch dispatch.** R 4.1.0 ships `bin/i386/` and `bin/x64/` (4.5.1 is x64-only), so "picks i386" was a fair guess. `0x403195`, straight after the prologue, unconditionally writes `x64`.

What is left is the hand-off: `bin/R.exe` re-spawns `bin\x64\R.exe` through msvcrt's `system()` (`""%s\bin\%s\R.exe"` in `.rdata`, `system` in the imports), and our probe passes `-e "cat(R.version$major, ".", R.version$minor, sep='')"` — embedded quotes, through a rebuilt command line. That is where I stopped, because I do not have a Windows box and I am not going to guess at a fix I cannot run.

So the narrowing is deliberate and argued rather than convenient: the failure this check exists for is a **libc mismatch**, which is this PR's entire subject and a Linux question. Windows ships one build per release and has no libc to mismatch. Gating the install path there on a probe that is demonstrably unreliable trades a real bug for a hypothetical one — which is your own point about a flaky verification step being worse than none, applied to the one platform where it was flaky. macOS keeps the check; the probe is known-good there, which is how the #99 path at `:598` came to be written that way.

One thing I want on the record rather than quietly fixed: that same `:598` short-circuit probes with `query_r_version` too, so on Windows an **already-installed** version fails the probe today. Pre-existing, not introduced here, not repaired here — it wants a Windows machine and its own change.

Thanks for tracing the #135 interaction; that was the second place I looked too, and I agree the worst case is duplicated work rather than corruption.

Rebase is a no-op — the branch is already on `bacdfe5`. Stack note unchanged: merges cleanly with #211/#213/#215, only the `## Unreleased` changelog anchor collides, so whichever lands second rebases one hunk. I re-verified that by merging all four just now.

nbafrank:
Merged — and thank you for root-causing it instead of letting a green check speak for itself. That was the whole question, and the answer turned out to be more interesting than a flake.

The finding that uvr has never probed a managed R on Windows (because `find_all()` takes the version from the directory name and never asks the binary) is worth having written down; it explains why adding the gate deleted a working 4.1.0 rather than catching a real problem. Scoping the check to macOS and Linux is the right call, and putting the analysis in the PR description means the next person doesn't have to reconstruct it from a thread.

Issue #120, opened 2026-06-17
bioc: fetch R↔Bioconductor release mapping from config.yaml instead of a hardcoded table


nbafrank:
Follow-up to #119. The R→Bioconductor release mapping in `bioc_release_for_r` (`crates/uvr-core/src/registry/bioconductor.rs`) is a hardcoded table that must be extended every R/Bioconductor release. #119 was a direct consequence: R 4.6 shipped, the table lacked it, and the fallback silently picked a release paired with the wrong R version.

**Proposal:** fetch Bioconductor's authoritative `r_ver_for_bioc_ver` map from https://bioconductor.org/config.yaml at resolve time (cached like other indices), and use it to pick the release for the active R version. Keep the hardcoded table as an offline fallback so resolution still works without network.

Benefits: the mapping never goes stale; new R/Bioc releases work without a uvr update. Mirrors how we already vendor-with-fallback for sysreqs (#103).

Considerations: one more network fetch at lock time (cacheable); config.yaml format stability; offline behavior must remain correct via the vendored table.

gdevenyi:
+1. Confirmed at `bioconductor.rs:22-34`: the hardcoded table has `_ => "3.23"` as fallback, silently mapping any future R (4.7, 4.8, 5.0) to Bioc 3.23. The `_` arm is deliberately identical to the `(4, 6)` arm per the "err toward latest API" policy at `:17-19`, but it means every new R release silently gets the newest-known Bioc until someone hand-edits this table. Same class of bug as #119 recurs every R release.

Fetching `config.yaml` at resolve time (cached) is the durable fix; the table should remain only as offline fallback.

Audit context: comprehensive codebase review.

nbafrank:
Implemented on main (bf35aa8): resolution now fetches bioconductor.org/config.yaml (cached on disk for 7 days), parses `r_ver_for_bioc_ver`, and picks the newest *released* Bioc paired with the active R — `devel_version` excluded, since a lockfile shouldn't pin packages built against a moving branch. Fallback chain: fresh cache → stale cache (offline) → the hardcoded table, which stays as the safety net. Live-verified: R 4.5 → 3.22, R 4.6 → 3.23. This ends the every-six-months staleness class (the R 4.5 row had been stale since October — exactly what PR #198 ran into). Ships in v0.4.4.

PR #217, opened 2026-08-01
fix: align vendored R 4.5 → Bioc 3.22 with the table's later-paired-release convention

gladkia:
Every other entry in the vendored `bioc_release_for_r` table maps R to its *later* paired Bioconductor release:

| R   | Earlier Bioc | Table entry |
|-----|-------------|-------------|
| 4.4 | 3.19        | 3.20        |
| 4.5 | 3.21        | ~~3.21~~ → 3.22 |
| 4.6 | 3.22        | 3.23        |

R 4.5 was the one outlier. This PR restores consistency.

**Note:** Since v0.4.4 the live `bioc-config.yaml` fetch already returns 3.22 for R 4.5 when a network is available. This fixes the offline / cold-cache fallback to match.

### Changes

- `bioc_release_for_r(4, 5)` returns `"3.22"` instead of `"3.21"`
- All in-tree tests that assert the R 4.5 static-table fallback updated to `"3.22"`
- Comment in `config_yaml_picks_latest_released_bioc_per_r` updated (table and live config now agree)

Replaces #198 (same fix, corrected rationale, rebased on v0.4.4, tls-repro.sh split out).

### DIFF
--- a/crates/uvr-core/src/registry/bioconductor.rs
+++ b/crates/uvr-core/src/registry/bioconductor.rs
@@ -37,7 +37,7 @@ const BIOC_CONFIG_TTL: Duration = Duration::from_secs(7 * 24 * 60 * 60);
 fn bioc_release_for_r(r_major: u64, r_minor: u64) -> &'static str {
     match (r_major, r_minor) {
         (4, 6) => "3.23",
-        (4, 5) => "3.21",
+        (4, 5) => "3.22",
         (4, 4) => "3.20",
         (4, 3) => "3.18",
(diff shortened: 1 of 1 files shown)


nbafrank:
Merged — thanks for splitting this out with the corrected rationale. The table's "later paired release" convention is the right justification, and R 4.5 was genuinely the lone outlier. Since v0.4.4 the live `config.yaml` fetch (#120) already returns 3.22 when a network is available, so this is squarely the offline-fallback fix, which is how the PR frames it. Exactly right.

Issue #201, opened 2026-07-29
rustls: "invalid peer certificate: UnknownIssuer" behind corporate TLS-inspecting VPN/proxy


gladkia:
## Symptom

Behind a corporate VPN or TLS-inspecting proxy, all `--bioc` package installs fail:

```
x ERROR  Failed to fetch Bioconductor index
  Network error: error sending request for url (https://bioconductor.org/packages/3.22/...)
  client error (Connect)
  invalid peer certificate: UnknownIssuer
```

P3M binary fetch also fails silently:

```
WARN  P3M CRAN binary index unavailable (Network error: error sending request ...)
```

Both curl and Python requests work fine with the same URLs.

## Root cause

Corporate TLS-inspecting VPNs/proxies replace the original TLS certificates with their own CA-signed certificates. The corporate Root CA is trusted by the OS system cert store (used by curl, Python, browsers) but NOT by rustls's webpki-roots bundle (which only contains public CAs).

Result: all HTTPS requests from uvr fail with UnknownIssuer when behind such a proxy.

## Reproduction

**With corporate VPN (SSL inspection):** disconnect VPN → `uvr lock` works. Reconnect → UnknownIssuer.

Note: uvr does not read `https_proxy` env var (reqwest built with `default-features = false`, no proxy feature). The failure occurs at the system network/TLS level when the corporate VPN replaces server certificates, not at the application proxy level.

**Self-contained reproduction script (no sudo, no Rust/cargo required):**

```bash
curl -fsSL "https://raw.githubusercontent.com/gladkia/uvr/fix/bioc-version-mapping-r45/repro/tls-repro.sh" -o /tmp/tls-repro.sh && bash /tmp/tls-repro.sh
```

Requirements: openssl, curl, python3 (standard on macOS and most Linux distros).

Expected output:
```
Test A (curl --cacert): ✅ SUCCESS  — tool trusts CA when told explicitly (like native-tls)
Test B (curl plain):    ❌ FAIL     — tool rejects CA when not in trust store (like rustls)
```

This demonstrates the core problem: rustls's webpki-roots bundle behaves like Test B (no custom CAs), while native-tls / system cert store behaves like Test A.

## Related

#200 — request to ship a native-tls binary as an alternative download (the fix for this issue)

## Fix options

(a) Fall back to system trust store when webpki-roots fails: `reqwest::ClientBuilder::with_native_roots()` (reqwest 0.12 API)

[...]

Issue #200, opened 2026-07-29
Release: provide native-tls binary as alternative download for enterprise environments


gladkia:
## Context

Enterprise environments often have internal TLS certificates not included in the rustls webpki-roots bundle (e.g. internal CDNs, package mirrors, Bioconductor's xsede.org CDN mirror). This causes "invalid peer certificate: UnknownIssuer" for Bioconductor CDN and custom package repositories.

## Request

Ship a native-tls release binary alongside the default rustls binary:

```
uvr-x86_64-unknown-linux-gnu.tar.gz          (current, rustls)
uvr-x86_64-unknown-linux-gnu-native-tls.tar.gz  (new)
```

Or provide a Cargo feature flag in the install script:

```
UVR_TLS=native curl ... | sh
```

## Workaround we're using

Rebuild uvr inside Docker with:

```bash
sed -i 's/rustls-tls/native-tls/g' Cargo.toml && cargo build --release
```

Note: the Bioconductor Docker base images (e.g. bioconductor/bioconductor_docker) already ship Rust, so no additional toolchain setup is needed in that context.

## Impact

Without native-tls, Bioconductor packages cannot be installed in environments where the xsede.org CDN cert is trusted by the system store but not by webpki-roots. This affects Bioc-heavy projects on non-standard cert infrastructure.

Related: #201 (root cause — rustls UnknownIssuer behind TLS-inspecting VPN)

nbafrank:
The underlying TLS issue is addressed on main via OS trust-store validation — details in https://github.com/nbafrank/uvr/issues/201#issuecomment-5136868941. If that verifies in your environment, this request becomes moot (the trust-store union covers the same CAs without doubling the release matrix), so holding off on a native-tls binary pending your retest.

gladkia:
Verified — see #201 for details. The OS trust-store approach in v0.4.4 covers our use case completely; a separate native-tls binary is not needed.

Closing as resolved by v0.4.4.

PR #218, opened 2026-08-01
docs: add repro/tls-repro.sh — self-contained TLS trust-store reproducer

gladkia:
Adds `repro/tls-repro.sh`, a self-contained script demonstrating the TLS trust-store problem addressed in #200/#201.

Requires only `openssl`, `curl`, and `python3` — no sudo, no Rust toolchain.

### Bugs fixed from the earlier version in #198

- **Trap double-quote**: `trap "... kill $SERVER_PID ..."` expanded `$SERVER_PID` at registration time (empty string), so the background server leaked on every run. Fixed by single-quoting the trap body.
- **Stray python3 invocation**: the script started a no-op `python3 -` process, captured its PID, then immediately overwrote `$SERVER_PID` with the real server. Fixed by writing the server to a temp file and running a single python3 process.

### Expected output

```
Test A (curl --cacert): ✅ SUCCESS — tool trusts custom CA
Test B (curl plain):    ❌ FAIL    — cert rejected (CA not trusted)
```

Related: #200 #201

### DIFF
--- a/repro/tls-repro.sh
+++ b/repro/tls-repro.sh
@@ -1,0 +1,114 @@
+#!/bin/bash
+# Reproduce the TLS trust-store problem: custom CA accepted by the OS cert
+# store but rejected by rustls's bundled webpki-roots.
+#
+# NO sudo, NO cargo/Rust required — uses openssl + curl + python3 only.
+#
+# Demonstrates:
+#   Test A (curl --cacert): ✅ tool trusts CA when told explicitly
+#   Test B (curl plain):    ❌ tool rejects CA when not in trust store
+#
+# This mirrors what happens behind a corporate TLS-inspecting VPN:
+#   VPN CA is in the OS cert store → curl/R/Python work (like Test A)
+#   rustls uses webpki-roots only, ignores OS store → uvr fails (like Test B)
+#
+# Note: uvr ≥ 0.4.4 resolves this by reading the OS trust store (#200/#201).
+# This script remains useful as a self-contained TLS-inspection reproducer.
+#
+# Requirements: openssl, curl, python3
+# Usage: bash tls-repro.sh
+
+set -e
+WORK=$(mktemp -d)
+SERVER_PID=""
+
+# Single-quote the trap so $WORK and $SERVER_PID are evaluated at exit time,
+# not at registration time (when SERVER_PID is still empty).
+trap 'rm -rf "$WORK"; [ -n "$SERVER_PID" ] && kill "$SERVER_PID" 2>/dev/null || true' EXIT
+
+echo "=== Requirements ==="
+for cmd in openssl curl python3; do
+  command -v "$cmd" &>/dev/null \
+    && echo "  ✅ $cmd" \
+    || { echo "  ❌ $cmd not found"; exit 1; }
+done
+echo ""
+
+echo "=== Step 1: Generate self-signed CA (not in webpki-roots or system store) ==="
+openssl genrsa -out "$WORK/ca.key" 2048 2>/dev/null
+openssl req -x509 -new -key "$WORK/ca.key" -out "$WORK/ca.crt" \
+  -days 1 -nodes -subj "/CN=test-corporate-CA" 2>/dev/null
+echo "Custom CA: $WORK/ca.crt  (CN=test-corporate-CA)"
+echo ""
+
+echo "=== Step 2: Generate server cert signed by custom CA ==="
+openssl genrsa -out "$WORK/server.key" 2048 2>/dev/null
+openssl req -new -key "$WORK/server.key" -out "$WORK/server.csr" \
+  -subj "/CN=127.0.0.1" 2>/dev/null
+openssl x509 -req -in "$WORK/server.csr" -CA "$WORK/ca.crt" -CAkey "$WORK/ca.key" \
+  -CAcreateserial -out "$WORK/server.crt" -days 1 \
+  -extfile <(echo "subjectAltName=IP:127.0.0.1") 2>/dev/null
+echo "Server cert: signed by test-corporate-CA"
+echo ""
+
+echo "=== Step 3: Start local HTTPS server (no sudo) ==="
+PORT=$(python3 -c "
+import socket
+s = socket.socket()
+s.bind(('', 0))
+p = s.getsockname()[1]
+s.close()
+print(p)
+")
+
+# Write the server script to a temp file so there is exactly one python3
+# invocation whose PID we capture.
+cat > "$WORK/server.py" << 'PYEOF'
+import http.server, ssl, sys
+
+work, port = sys.argv[1], int(sys.argv[2])
+
+class H(http.server.BaseHTTPRequestHandler):
+    def do_GET(self):
+        self.send_response(200)
+        self.end_headers()
+        self.wfile.write(b"Package: BiocStyle\nVersion: 2.38.0\n")
+    def log_message(self, *a): pass
+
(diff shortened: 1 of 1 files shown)


nbafrank:
Thanks for splitting this out. I ran it three times in a sandbox and confirmed two of the three issues from the earlier review are fixed: the trap is single-quoted now so `$SERVER_PID` defers to signal time (verified zero lingering `server.py` processes after consecutive runs), and the port-discovery/server split removed the dead first `python3` invocation. Clean exits, no leftover temp dirs.

The third is still open, and it's the one that matters for what this script is *for*: Test A uses `curl --cacert "$WORK/ca.crt"`, an explicit CA override — it never installs the CA into the OS trust store and never invokes `uvr` at all. So it demonstrates a generic curl/TLS fact rather than the #201 claim that uvr reads the OS store. Your PR body doesn't claim otherwise, so this isn't a discrepancy — but to be the reproducer for #201 it needs `security add-trusted-cert` / `update-ca-certificates` (or a documented manual step), then a plain `curl` with no flag, then `uvr` itself against the local server.

Separately: v0.4.4 shipped the actual fix (OS trust store validation alongside webpki roots). If you can retest your original corporate-proxy failure against 0.4.4 or later, that closes #201 and #200 — I can't reproduce a TLS-inspecting proxy here, so your confirmation is the last step.

gdevenyi:
Drive-by — @nbafrank's third point is the one I could actually help with, so I built the reproducer it describes and ran it. Two useful results, one of which is a bug nobody has filed.

**You do not need a TLS-inspecting proxy to reproduce this.** That is the blocker in the thread ("I can't reproduce a TLS-inspecting proxy here"), and it turns out to be avoidable: point the CDN hostname at a local TLS server whose certificate is signed by a CA you control, via `/etc/hosts`. That is the same trust question a corporate proxy poses — a cert chaining to a root the OS knows and webpki does not — without any proxy. It runs in a stock `debian:12` container in about 20 seconds.

```sh
openssl req -x509 -newkey rsa:2048 -nodes -keyout ca.key -out ca.crt -days 1 \
  -subj "/CN=uvr-repro-test-CA"
openssl req -newkey rsa:2048 -nodes -keyout srv.key -out srv.csr -subj "/CN=cdn.posit.co"
printf 'subjectAltName=DNS:cdn.posit.co\n' > ext.cnf
openssl x509 -req -in srv.csr -CA ca.crt -CAkey ca.key -CAcreateserial \
  -out srv.crt -days 1 -extfile ext.cnf

# serve /r/versions.json over TLS on :443, then
echo "127.0.0.1 cdn.posit.co" >> /etc/hosts
# ... probe, install the CA, probe again
cp ca.crt /usr/local/share/ca-certificates/ && update-ca-certificates
```

`uvr r list --all` is the ideal probe: one GET to `https://cdn.posit.co/r/versions.json`, no R install needed, no project needed.

**Result 1 — the fix works.** Against the shipped v0.4.4 binary, `uvr` and a bare `curl` (no `--cacert`, which was @nbafrank's objection) flip together:

```
== BEFORE installing the CA into the OS trust store ==
  uvr r list --all         FAILS
  curl (no flag)           TLS REJECTED

== AFTER  update-ca-certificates ==
  uvr r list --all         OK      <- server logged: "GET /r/versions.json HTTP/1.1" 200
  curl (no flag)           OK
```

So uvr does read the OS trust store, and now there is a test that demonstrates it rather than demonstrating a generic curl fact. `Cargo.toml` agrees — `reqwest` carries `rustls-tls-native-roots` and `ureq` carries `platform-verifier` — but the point of this script is not to trust that.

**Result 2 — and this is the part I would file: when TLS fails, uvr does not say so.** Look at what the "BEFORE" row actually prints:

```
x ERROR  Failed to fetch available R versions: Network error: error sending request for
url (https://cdn.posit.co/r/versions.json)
```

No "certificate", no "UnknownIssuer", nothing. That is indistinguishable from being offline, from DNS failing, from the CDN being down. The cause is `crates/uvr-core/src/error.rs:43`:

```rust
#[error("Network error: {0}")]
Network(#[from] reqwest::Error),
```

`{0}` on a `reqwest::Error` renders only the outermost layer. The real reason — rustls' `InvalidCertificate(UnknownIssuer)` — is sitting in the `source()` chain and never gets walked, so it never reaches the user.

I would argue this is *why* #201 took the shape it did. The original report's `"invalid peer certificate: UnknownIssuer"` had to come from somewhere other than uvr's own error output, and anyone hitting it today on a version *without* the native-roots fix would see only `Network error: error sending request` and reasonably conclude their network was broken. Walking the source chain in the `Display` impl is a small change and turns this class of report from "uvr does not work at my company" into a self-diagnosing message.

Happy to file that separately if you would both rather keep this PR to the script. And @nbafrank, for the retest you asked for: the above is a standalone answer to "does 0.4.4 validate against the OS store" that does not depend on @gladkia having their original corporate network to hand — though their confirmation is obviously still the real-world close for #201/#200.

gladkia:
@gdevenyi — the `/etc/hosts` + `update-ca-certificates` reproducer is a much better approach than the `--cacert` script, and your v0.4.4 result is the real-world confirmation @nbafrank needed.

I can also confirm from the corporate-VPN side: `uvr lock` with a Bioconductor dependency on Intel Mac behind a TLS-inspecting VPN, v0.4.4 — resolves cleanly with no TLS errors.

gdevenyi:
Correcting my own comment above: **the diagnosis was wrong, and the bug is smaller and easier to fix than I said.**

I blamed `error.rs:43` for swallowing the TLS cause. That is not it. The error chain is intact — I just happened to probe with the one command in the codebase that throws it away.

Same reproducer, same container, same untrusted CA, only the uvr subcommand changed:

```
uvr r list --all
  x ERROR  Failed to fetch available R versions: Network error: error sending
    request for url (https://cdn.posit.co/r/versions.json)          <- no cert mention

uvr r install 4.5.1
  certificate: UnknownIssuer                                        <- says it plainly
```

The difference is one line at each call site:

```rust
// r_cmd/list.rs:18 — flattens the cause into a string, chain lost
.map_err(|e| anyhow::anyhow!("Failed to fetch available R versions: {e}"))?

// r_cmd/install.rs:51 — preserves it, so the printer walks it
.context("R installation failed")?
```

`{e}` renders only `UvrError::Network`'s `Display`, which is `Network error: {0}`, and `{0}` on a `reqwest::Error` is just the outermost layer. Everything below it is dropped right there. `.context()` keeps the source chain, so the printer reaches rustls' `UnknownIssuer`.

And this is very much the exception rather than the rule — I counted:

```
.context("...")                              86 call sites
map_err(|e| anyhow!("...{e}"))                2 call sites   <- both in r_cmd/list.rs (15, 18)
```

Two out of eighty-eight, and I picked one of the two as my probe because `r list --all` was the cheapest way to force a single HTTPS GET. So my "uvr does not say so" claim holds for `uvr r list`, and **not** for `uvr add`, `uvr sync` or `uvr r install` — which is where a user behind a corporate proxy actually hits it first. #201's reporter would have seen the real message.

That makes it a two-line fix in `list.rs` rather than anything structural, and it does not need its own issue unless you want one — happy to send it as a small PR.

Independent corroboration that the good path works, from an unrelated CDN flake on #232 today:

```
  Network error: error sending request for url (…musllinux_1_2…)
  client error (Connect)
  Connection reset by peer (os error 104)
```

The rest of my comment stands: the `/etc/hosts` + local-CA reproducer works and needs no proxy, and it confirms v0.4.4 validates against the OS trust store.

Issue #202, opened 2026-07-29
sysreqs: Posit API returns 500 for Bioconductor packages — noisy warnings in Bioc-heavy projects


gladkia:
## Symptom

During `uvr sync` on a Bioconductor-heavy project, every Bioc package triggers a warning:

```
WARN  Posit sysreqs API returned 500 Internal Server Error for SummarizedExperiment
WARN  Posit sysreqs API returned 500 Internal Server Error for BiocParallel
WARN  Posit sysreqs API returned 500 Internal Server Error for MultiAssayExperiment
[... repeated for every Bioc package ...]
```

In a project with ~100 Bioc transitive dependencies, this generates ~100 WARN lines that obscure real warnings.

## Reproduction

```bash
mkdir /tmp/uvr-bioc-test && cd /tmp/uvr-bioc-test
uvr init test --r-version 4.5

cat >> uvr.toml << 'EOF'

bioc_version = "3.22"

[dependencies.SummarizedExperiment]
bioc = true
EOF

uvr sync 2>&1 | grep "sysreqs API"
# → many lines of "WARN  Posit sysreqs API returned 500 Internal Server Error for ..."
```

## Root cause

The Posit sysreqs API (https://sysreqs.r-pkg.org/) only covers CRAN packages. Bioconductor packages return HTTP 500. This is expected API behaviour, but uvr surfaces it as a per-package WARN rather than a single summary.

## Impact

- Bioc packages with system dependencies (rhdf5 → libhdf5, BiocParallel → libomp, RcppParallel → libomp) are silently skipped — no sysreqs hint is emitted for them
- Users see ~100 identical WARN lines that make it hard to spot real warnings
- The log noise is particularly bad in CI where log size matters

## Suggested fix

Suppress per-package 500 errors and emit a single summary at the end of sysreqs detection:

```
Note: sysreqs API does not cover Bioconductor packages (N packages skipped).
      If Bioc packages fail to compile, install system libraries manually.
```

Optionally, hardcode sysreqs for the most common Bioc packages with system deps:
- `rhdf5`, `HDF5Array` → `libhdf5-dev`
- `BiocParallel`, `RcppParallel`, `fgsea` → `libomp-dev`
- `RCurl` → `libcurl4-openssl-dev`
- `XML` → `libxml2-dev`

## Context

[...]

PR #219, opened 2026-08-01
fix(sysreqs): don't report a successful local check as skipped

pat-s:
## The problem

Every Alpine build prints a warning that is false:

```
 ! WARN  System dependency check skipped on alpine-3.23.5
  Posit's sysreqs catalog doesn't cover this distribution, and the local fallback had no applicable rules.
```

Both halves are wrong.

**The catalog does cover Alpine.** In the vendored snapshot, 106 of 131 rule files carry alpine constraints, and 99 of the 111 alpine constraints have no `versions` key at all. What doesn't cover Alpine is Posit's *HTTP API*, which returns `{"code":14,"error":"Unsupported system"}`. The message blamed the wrong component.

**And the check had actually succeeded.** When the API declines a distro, `check_system_deps` routes every package through the vendored rules, which resolve Alpine correctly (`curl` → `curl-dev`, `openssl-dev`; `xml2` → `libxml2-dev`, including the `3.23.5` → `3.23` truncation). Those go through `filter_missing`, and on a pre-provisioned image everything is already installed, so `missing` comes back empty. `SysReqsCheck` had no field distinguishing "the fallback ran and found nothing missing" from "the fallback could not check", so the branch reported a successful check as a skipped one.

### Why it started appearing after #208

Before #208, `SystemRequirements` was `None` for every CRAN package, so `any_has_sysreqs` was false and this branch never fired — as #208's own commit message notes. By populating the field from the tarball, #208 flipped that gate true while leaving the `missing.is_empty()` conflation intact. The fix surfaced a latent bug.

## The change

`SysReqsCheck` gains `local_resolved`, counting packages the vendored rules resolved to at least one system package. It is incremented in `check_pkg_local` *before* `filter_missing` shells out, so it is independent of what happens to be installed.

Both warning branches — `lookup_failed` and `unsupported_distro` — then gate on `local_check_incomplete()`, which compares `local_resolved` against the number of packages that actually declare `SystemRequirements`. A per-package comparison rather than `== 0` matters: the any-package form would silence the warning for a whole sync as soon as one package resolved, and since `curl`/`openssl`/`xml2` resolve nearly everywhere, that would have made the warning unreachable.

The skipped-check message now names the API rather than the catalog.

## Verification

In `build-env-alpine:3.23` (`VERSION_ID=3.23.5`), with `harfbuzz-dev`/`fribidi-dev` installed so every `textshaping` requirement resolves and is present:

- before: `! WARN  System dependency check skipped on alpine-3.23.5`
- after: silent

Positive control, same package with those two absent — must still fire, and does:

```
 ! WARN  Missing system dependencies for 1 package(s)
  textshaping needs: fribidi-dev, harfbuzz-dev
```

Tests: 492 → 496.

### DIFF
--- a/crates/uvr-core/src/sysreqs.rs
+++ b/crates/uvr-core/src/sysreqs.rs
@@ -586,7 +586,9 @@ mod tests {
         let mut out = SysReqsCheck::default();
         let pkg = PackageSysReqQuery {
             name: "xml2".to_string(),
-            system_requirements: Some("libxml2: libxml2-dev (deb), libxml2-devel (rpm)".to_string()),
+            system_requirements: Some(
+                "libxml2: libxml2-dev (deb), libxml2-devel (rpm)".to_string(),
+            ),
             bioc: false,
         };
         check_pkg_local(&mut out, &pkg, "alpine-3.23.5");
--- a/crates/uvr-core/src/sysreqs.rs
+++ b/crates/uvr-core/src/sysreqs.rs
@@ -240,12 +240,19 @@ pub struct SysReqsCheck {
     /// Missing system packages keyed by R package name.
     pub missing: HashMap<String, Vec<SysReq>>,
     /// Set when the Posit API reported the distro as unsupported.
-    /// When true, `missing` is not authoritative — the check was skipped.
+    /// When true, the API contributed nothing — but the vendored local
+    /// rules still ran, so `missing` may well be authoritative. Read this
+    /// together with `local_resolved` before claiming the check was skipped.
     pub unsupported_distro: bool,
     /// Set when at least one API lookup failed outright (network/HTTP/parse,
     /// #148). Affected packages were checked against the vendored local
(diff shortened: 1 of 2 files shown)


nbafrank:
Merged. Verified the bug is real on main — both warning gates fired on a bare `check.missing.is_empty()`, which can't distinguish "the local rules resolved everything and it's all installed" from "we never checked." Counting resolutions in `check_pkg_local` *before* `filter_missing` runs is the right call: it makes the signal independent of what happens to be installed on the host.

One subtlety worth knowing since it isn't in the PR body: Bioconductor packages skip the API entirely (#202) and go straight to `check_pkg_local`, so they contribute to `local_resolved`/`pkgs_with_sysreqs` too. That's correct — an unmatched Bioc `SystemRequirements` is a genuine coverage gap — but it means the warning can now fire on a sync that never touched the API at all.

PR #220, opened 2026-08-01
perf(sysreqs): one Posit API request per sync instead of one per package

pat-s:
Stacked on #219 — review that first; this branch contains its commits.

## The change

`check_system_deps` queried Posit's sysreqs API once per package, breaking out to the vendored rules on the first `Unsupported system` response. The API also serves a whole distribution in one call, so this fetches `?all=true&distribution=…&release=…` once per sync (~150 KB, ~1125 entries for ubuntu 22.04) and looks each package up in it.

Measured on `ubuntu:22.04` with a 32-package resolution: **32 requests → 1**.

## Equivalence

The resolved system-package sets are byte-identical before and after — `diff` over the deduplicated sorted sets is empty, and the aggregate `Install with: apt-get install -y …` line matches exactly.

**One visible output change, worth calling out.** The old per-package endpoint returns the *recursive dependency closure* tagged by `name`, and the old code ignored `req.name`, attributing every dependency to whichever package was queried — so `pkgname=gert` reported `credentials`, `curl` and `openssl`'s requirements as `gert`'s. The batched index attributes each to its own package. The aggregate set is unchanged because `check_system_deps` always receives the full transitive install set.

## Two things this deliberately preserves

**Malformed responses stay distinguishable from empty ones.** `parse_sysreqs_index` returns `Result`, and a JSON failure yields `LookupFailed`, not `Supported(empty)`. An infallible version would make a truncated 200 response look identical to "no package on this distro needs anything" — on every distro, silently — and would regress against the `resolve_system_deps` it replaces.

**All-Bioconductor syncs still make zero API requests.** The fetch is guarded on at least one non-Bioc package being present. Without that guard, batching would have quietly dropped the #202 guarantee and let a network hiccup flag `lookup_failed` on a sync that never needed the API.

## Known tradeoff

A single transient API failure now downgrades the whole sync to the vendored rules, where the old loop kept querying for the remaining packages. That is inherent to fetching once, and the local rules are a complete fallback.

## Structure

The per-package decision is extracted into a synchronous `apply_index`, so the branch selection is testable without network. Four tests cover unsupported-distro, index-hit, index-miss and mixed Bioc/CRAN, asserting on `local_resolved` rather than `missing` so they hold regardless of what is installed on the test host.

Tests: 496 → 504.

### DIFF
--- a/crates/uvr-core/src/sysreqs.rs
+++ b/crates/uvr-core/src/sysreqs.rs
@@ -3,7 +3,6 @@ use std::collections::HashMap;
 use serde::Deserialize;
 use tracing::{debug, warn};
 
-use crate::error::Result;
 use crate::sysreqs_rules;
 
 /// A resolved system dependency with its apt/rpm package name.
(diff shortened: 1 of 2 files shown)


nbafrank:
Merged. I verified the batch endpoint live before taking this — `?all=true&distribution=ubuntu&release=22.04` returns ~150 KB in the shape `PpmSysreqsResponse` expects, and Alpine still returns the same `Unsupported system` body, so unsupported-distro detection survives batching.

The attribution fix riding along is a real bug catch: `?pkgname=gert` returns `credentials`/`curl`/`gert`/`openssl` as four separately-named entries, and main had no `name` field at all, so all four were credited to `gert`. It washed out in the aggregate set, but it's right to fix.

Your disclosed tradeoff is accurate and worth restating for anyone reading later: a transient API failure now degrades the whole sync to local rules rather than just the unlucky package. That's inherent to 32-requests-to-1, the local fallback is complete, and #219 now tells the user when that fallback left gaps — good sequencing.

Note main moved: #214 landed alongside this and I re-applied its `catalog_alias` call inside your new `fetch_sysreqs_index` (it had patched the per-package function you replaced), so the RHEL-family aliasing now covers the batched request.

PR #221, opened 2026-08-01
feat(sysreqs): run the rules' pre_install/post_install commands

pat-s:
Stacked on #220 — review that first; this branch contains its commits.

## The gap

The sysreqs rules carry setup commands — 63 of the 131 vendored rule files have `pre_install`, 2 have `post_install` — and uvr dropped all of them. The build script never emitted them into the generated table, and `PpmRequirementDetail` deserialized only `packages`, discarding `install_scripts`/`pre_install`/`post_install` from the API response.

So `--install-system-deps` on Rocky 9 would try to install `gdal3.4-devel` from repos that were never enabled, and `rJava` never got `R CMD javareconf`.

Both paths had to change. Fixing only the local resolver would look right and do nothing where it matters: on RHEL and its rebuilds — exactly the distros carrying the interesting commands — the API answers, so `check_pkg_local` is never reached.

## The change

`DependencyStatic` and `resolve_local` carry the commands; `PpmRequirementDetail` deserializes them; `SysReqsCheck` collects them, deduplicated in stable order across matched rules. `run_rule_commands` runs `pre_install` before the batched package install and `post_install` after.

Collection is gated on something actually being missing, so a fully-provisioned machine never re-enables EPEL on every sync.

## Security posture — the part worth your attention

This extends uvr's executed surface from package-manager invocations it builds itself to **shell strings from a data file, run as root**. That includes `add-apt-repository -y ppa:ubuntugis/ppa` and `rpm -q epel-release || yum install -y https://…`. `pak` and `rig` already execute the same rule data, which is the precedent this follows.

Deliberate choices:

- **`sh -c`, not a direct spawn** — several commands rely on `||`.
- **Nothing runs without opt-in.** The only call sites sit in the `(true, Some(..))` arm behind `--install-system-deps` / `UVR_INSTALL_SYSREQS`, inside `!missing.is_empty()`.
- **The invoking user's `$PATH` never reaches a root shell.** `pre_install` passes no `PATH` override at all, so sudo's `secure_path` applies unchanged. Only `post_install` needs R, and it prepends the R bin directory to a fixed `ROOT_SAFE_PATH` mirroring the stock `secure_path`. Handing `$PATH` through would let root resolve `rpm` or `curl` from a user's conda or `~/.local/bin` — and `rust.json` pipes a remote script through `curl`.
- **Consent covers what actually runs.** The full command list prints before the confirmation, on stderr alongside the prompt, and unconditionally so non-interactive runs log it too.
- **Asymmetric failure handling.** `pre_install` bails — a repo that could not be enabled makes the following install fail anyway. `post_install` warns and continues, naming the command to re-run: the packages are already installed by then, and a failing `javareconf` should not turn a successful install into a failed sync.

## Verification, on stock `rockylinux:9` with EPEL confirmed absent

```
Running: dnf install -y epel-release
Running: dnf install -y dnf-plugins-core
Running: dnf config-manager --set-enabled crb
Running: dnf install -y … gdal3.4-devel geos-devel proj-devel …
```

gdal/geos/proj then resolve *from* `epel`/`crb` in the transaction. `rJava` gets `R CMD javareconf`. With the full transitive set pre-installed, zero setup commands run.

## Known limitation

Deduplication keeps the first occurrence, which assumes rule commands are order-independent across rules. On EOL Ubuntu 16.04 a `sf` + `gert` sync can land a PPA after the last `apt-get update`. Documented in the code; 18.04+ `add-apt-repository` self-updates.

Tests: 504 → 511.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,31 @@ release page on GitHub. Issue numbers reference https://github.com/nbafrank/uvr/
 
 Pure tracking section — fixes and small features land here between tags.
 
+- **`--install-system-deps` now runs the setup commands the sysreqs rules
+  carry**, so it works on distros where the packages do not exist until a
+  repository is enabled.
+  63 of the 131 vendored rule files have a `pre_install` and 2 have a
+  `post_install`, and uvr dropped all of them: on Rocky 9 it tried to
+  install `gdal3.4-devel` from repos that were never enabled, and `rJava`
+  never got its `R CMD javareconf`.
+  `pre_install` now runs before the batched package install and
+  `post_install` after, both only when something is actually missing, so a
+  fully provisioned machine does not re-enable EPEL on every sync.
+  Nothing runs without `--install-system-deps` / `UVR_INSTALL_SYSREQS=1`.
+
+- **The consent block says where each root command comes from and what it
+  does.**
+  These are shell strings run as root, and two of them are worth reading
+  before approving: they arrive either from uvr's vendored
+  `r-system-requirements` snapshot or from Posit's API fetched during that
+  sync, and some of them fetch and execute remote code, install from a
+  URL, disable signature checking, or add a third-party repository.
+  Each command is now listed with both, on stderr alongside the prompt and
+  unconditionally, so redirecting stdout to a build log cannot separate
+  the consent from what it covers.
+  The block also states that an unannotated command is one uvr matched no
+  pattern in rather than one it vetted.
+
 ## v0.4.6 (2026-08-10)
 
 Standalone scripts land: a `.R` file can now declare its own dependencies
--- a/crates/uvr-core/build.rs
+++ b/crates/uvr-core/build.rs
@@ -18,10 +18,19 @@ struct Rule {
 }
 
 #[derive(Debug, Deserialize)]
+struct Command {
+    command: String,
+}
+
+#[derive(Debug, Deserialize)]
 struct Dependency {
     packages: Vec<String>,
     #[serde(default)]
     constraints: Vec<Constraint>,
+    #[serde(default)]
+    pre_install: Vec<Command>,
+    #[serde(default)]
+    post_install: Vec<Command>,
 }
 
 #[derive(Debug, Deserialize)]
@@ -69,7 +78,7 @@ fn main() {
          // Do not edit by hand.\n\n",
     );
     out.push_str("pub struct ConstraintStatic {\n    pub os: Option<&'static str>,\n    pub distribution: Option<&'static str>,\n    pub versions: &'static [&'static str],\n}\n\n");
-    out.push_str("pub struct DependencyStatic {\n    pub packages: &'static [&'static str],\n    pub constraints: &'static [ConstraintStatic],\n}\n\n");
+    out.push_str("pub struct DependencyStatic {\n    pub packages: &'static [&'static str],\n    pub constraints: &'static [ConstraintStatic],\n    pub pre_install: &'static [&'static str],\n    pub post_install: &'static [&'static str],\n}\n\n");
     out.push_str("pub struct RuleStatic {\n    pub name: &'static str,\n    pub patterns: &'static [&'static str],\n    pub dependencies: &'static [DependencyStatic],\n}\n\n");
     writeln!(out, "pub static RULES: &[RuleStatic] = &[").unwrap();
     for (name, rule) in &rules {
@@ -107,6 +116,16 @@ fn main() {
                 writeln!(out, "] }},").unwrap();
             }
             writeln!(out, "                ],").unwrap();
+            write!(out, "                pre_install: &[").unwrap();
+            for c in &dep.pre_install {
+                write!(out, "{:?}, ", c.command).unwrap();
+            }
+            writeln!(out, "],").unwrap();
+            write!(out, "                post_install: &[").unwrap();
+            for c in &dep.post_install {
+                write!(out, "{:?}, ", c.command).unwrap();
+            }
+            writeln!(out, "],").unwrap();
             writeln!(out, "            }},").unwrap();
         }
         writeln!(out, "        ],").unwrap();
--- a/crates/uvr-core/src/installer/binary_install.rs
+++ b/crates/uvr-core/src/installer/binary_install.rs
@@ -1111,7 +1111,7 @@ mod tests {
         let pkgs = crate::sysreqs_rules::resolve_local(&sys_reqs, "alpine", "3.24.1");
         for expected in ["gdal-dev", "geos-dev", "proj-dev", "sqlite-dev"] {
             assert!(
-                pkgs.iter().any(|p| p == expected),
+                pkgs.packages.iter().any(|p| p == expected),
                 "expected {expected}, got {pkgs:?}"
             );
         }
--- a/crates/uvr-core/src/sysreqs.rs
+++ b/crates/uvr-core/src/sysreqs.rs
@@ -260,9 +260,29 @@ struct PpmRequirement {
 }
 
 #[derive(Debug, Default, Deserialize)]
+struct PpmCommand {
+    #[serde(default)]
+    command: String,
+}
+
+#[derive(Debug, Default, Deserialize)]
 struct PpmRequirementDetail {
     #[serde(default)]
     packages: Vec<String>,
+    #[serde(default)]
+    pre_install: Vec<PpmCommand>,
+    #[serde(default)]
+    post_install: Vec<PpmCommand>,
+}
+
+/// One package's entry in the batched sysreqs index: its distro packages
+/// plus any setup commands the API attaches to them (e.g. enabling EPEL
+/// before a `dnf install`, or `R CMD javareconf` after one).
+#[derive(Debug, Default)]
+struct SysReqIndexEntry {
+    packages: Vec<SysReq>,
+    pre_install: Vec<String>,
+    post_install: Vec<String>,
 }
 
 /// Index a batched (`all=true`) sysreqs response by R package name.
@@ -277,8 +297,8 @@ struct PpmRequirementDetail {
 /// `LookupFailed`; keep that.
 fn parse_sysreqs_index(
     body: &str,
-) -> std::result::Result<HashMap<String, Vec<SysReq>>, serde_json::Error> {
-    let mut index: HashMap<String, Vec<SysReq>> = HashMap::new();
+) -> std::result::Result<HashMap<String, SysReqIndexEntry>, serde_json::Error> {
+    let mut index: HashMap<String, SysReqIndexEntry> = HashMap::new();
     let response: PpmSysreqsResponse = serde_json::from_str(body)?;
     for req in response.requirements {
         if req.name.is_empty() {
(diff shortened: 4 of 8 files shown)


nbafrank:
Reviewed with the scrutiny root-executing code deserves, and the engineering holds up well. Verified directly: no injection surface (the emitted commands are always static strings from the rule table — `SystemRequirements` text is only ever a `RegexSet` match trigger, never interpolated), `needs_sudo` correctly mirrors `pick_sysreqs_installer` so the "sudo needed but missing" state is unreachable, the full plan is disclosed *before* `confirm_sysreqs_install` including on non-TTY runs, and the `pre_install`/`post_install` PATH split is genuinely careful — `pre_install` deferring to sudo's own `secure_path` and `post_install` building from a fixed root-safe PATH plus the resolved R's parent. I also noticed you caught and fixed the PATH regression yourself mid-PR (204cfd8/0fd9d4d); that's the diligence this feature needs.

**One thing to fix before merge**, and it's disclosure rather than code:

The commands executed as root come from *two* provenances, and the consent prompt doesn't distinguish them. Vendored rules are compile-time constants from the audited `vendor/r-system-requirements` snapshot — but `SysReqIndexEntry` also carries `pre_install`/`post_install` straight out of the live Posit API response, and `apply_index` funnels both into the same `SysReqsCheck` fields, so `run_rule_commands` can't tell them apart. On any API-covered distro (Ubuntu, the RHEL family — i.e. most users), a meaningful share of what gets run as root was fetched over the network at sync time. I confirmed this is live, not theoretical: `?all=true&distribution=rockylinux&release=9` returns real `pre_install` entries like `dnf install -y epel-release` for `sf`/`terra`.

TLS verification is intact and pak/rig set the precedent, so I'm not asking you to drop the API path — just to label it. Something like tagging each command with its source when it's collected and printing "(from Posit's sysreqs API)" vs "(from uvr's vendored rules)" in the disclosure block. Someone approving root execution should be able to see which half is which.

Everything else is approved. Rebase note: #219, #220 and #214 have all landed on main since you filed, and #214 moved the RHEL aliasing into `fetch_sysreqs_index`.

gdevenyi:
Drive-by while I was in the sysreqs code for #211 — I went looking for evidence on @nbafrank's provenance point and found something that I think sharpens it rather than settling it. Not an objection to the feature; the disclosure design is the right instinct.

**Today the two provenances agree exactly.** I diffed the live API against the vendored snapshot for `rockylinux`/`9` — every `pre_install`/`post_install` command the API returns is already in `vendor/r-system-requirements`:

```
API (rockylinux/9) pre/post commands:                6
  [pre_install]  dnf install -y epel-release
  [pre_install]  dnf install -y dnf-plugins-core
  [pre_install]  dnf config-manager --set-enabled crb
  [pre_install]  yum install -y epel-release
  [pre_install]  curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- --no-modify-path -y
  [post_install] R CMD javareconf

In API but NOT in the vendored snapshot:             0
```

So the label would currently be accurate and, for every command, redundant. That is worth knowing before building it: it makes the label a **drift detector** — its value is the day the two disagree, not today.

**But look at the fifth line.** That is `curl | sh` as root, and it is in *both* halves. Which means a disclosure that distinguishes only by provenance would print:

```
curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- --no-modify-path -y
   (from uvr's vendored rules)
```

— true, reassuring, and pointed at the wrong axis. The vendored snapshot is audited in the sense that a human read the *rule*; it is not a guarantee about what the rule *does at runtime*. Across all 34 distinct pre/post commands in the snapshot, the ones I would want a user to actually see before approving are:

| what it does | example |
|---|---|
| fetches and executes remote code | `curl -sSf https://sh.rustup.rs \| sh` (rust.json) |
| downloads a vendor package and installs it | `curl -fsSL -o /tmp/google-chrome.deb https://dl.google.com/…` then `apt-get install -y /tmp/google-chrome.deb` (chrome.json) |
| **disables signature checking** | `dnf install -y --nogpgcheck https://mirrors.rpmfusion.org/…` (libavfilter.json) |
| adds a third-party repo to the system | `add-apt-repository -y ppa:ubuntugis/ppa`, `ppa:cran/libgit2`, `ppa:xtradeb/apps` |

Every one of those is vendored. So I would suggest the provenance tag @nbafrank asked for **and** that the disclosure not imply vendored means safe — the two labels answer different questions, and only the second is the one the user is actually consenting to. `--nogpgcheck` in particular seems worth surfacing on its own terms.

To be clear on scope: none of this is introduced by this PR. These commands exist in the snapshot on main today; what changes here is that uvr starts *running* them, which is exactly why it is the right moment to say so. And @nbafrank's verification that the command strings are never interpolated from `SystemRequirements` text is the property that makes this a disclosure problem rather than an injection one — I re-checked that and agree.

One unrelated thing worth knowing, since it lands in the same code: `pick_sysreqs_installer`'s unconditional `apt-get` fallback is wrong on zypper/pacman/yum-only hosts, and it reaches users as advice today. I filed it as #226 with the openSUSE reproduction from #211's distro suite. It does not affect `run_rule_commands` — the rule commands carry their own package manager — but if you touch that function for `needs_sudo`, the two are adjacent.

nbafrank:
@pat-s — checking in on this one. It's the last piece of the sysreqs work still outstanding, and I'd like it in the next release rather than letting it go stale.

The only thing I asked for is small: label each command's provenance in the consent block, so a user approving root execution can see which commands come from uvr's vendored `r-system-requirements` snapshot and which were fetched from Posit's API at sync time. Everything else — the sudo gating, PATH isolation, injection surface, consent ordering, failure handling — checked out.

Worth noting @gdevenyi found that today the two sources agree exactly (every `pre_install`/`post_install` the API returns for rockylinux/9 is already in the vendored snapshot), so the labelling is about being honest with the user rather than about a live discrepancy.

If you'd rather not spend more time on it, say so and I'll add the labelling myself and merge — the feature is good and I don't want it lost over a one-line disclosure change. Also note main has moved a fair bit (v0.4.5 and v0.4.6 both shipped), so a rebase is probably wanted either way.

pat-s:
You can always do a rebase yourself if you want to get something in.
I'll rebase myself then now.

pat-s:
Rebased onto main (v0.4.6), and both provenance points are addressed.

## Rebase

Four conflicts, three of them substantive rather than textual:

- **`SysReqsCheck`** — #214's `local_unresolved` and this branch's `pre_install`/`post_install` are both additive; kept both.
- **Setup vs. index refresh** — main grew `run_sysreqs_refresh()` inside the same arm this branch adds `run_rule_commands(pre_install)` to, so the merge had to pick an order. Refresh goes **first**: `gdal.json`'s Ubuntu `pre_install` opens with `apt-get install -y software-properties-common`, which needs lists a fresh container doesn't have. The rules re-refresh themselves where their own repo additions need it — that same rule ends with `apt-get update` after the PPA. Reasoning is in the code, since it isn't obvious from either side alone.
- **`install_hint`** — main's `sysreqs_install_hint` handles the "no known package manager" case by naming the packages instead of a command. This branch's two hint arms predated it and hardcoded `Install with: {cmd}`, which would have printed a bare `Install with:` on a host with no recognised manager. Both arms now go through the helper, with the setup commands still printed around it.
- Plus `install_env` (#232's `DEBIAN_FRONTEND`) merged with the `info_err` move, and the `resolve_local` signature change propagated into tests main added meanwhile.

`#226` is already fixed on main — `pick_sysreqs_installer` goes through `PackageManager::detect()` now, so the `apt-get` fallback @gdevenyi flagged is gone. Nothing left to do there on this branch.

## Disclosure

@nbafrank — provenance is now carried rather than reconstructed. `SetupCommand` pairs the shell string with a `CommandSource`, `check_pkg_local` tags `Vendored`, `apply_index` tags `PositApi`.

One thing that fell out of @gdevenyi's measurement: since the two catalogs agree exactly today, most commands arrive from *both*, and crediting whichever was seen first would have told users "vendored" about commands the API also sent — reassurance that isn't true of the second half. So `push_setup_command` merges the tag into `Both` on dedup. That makes the drift-detector framing literal: a single-sourced command is one the other catalog does not carry.

@gdevenyi — agreed that provenance alone points at the wrong axis, and the `curl | sh` example is the proof. Vendored means a human reviewed the *rule*; it says nothing about what the rule does when run. So there's a second, independent annotation for behaviour, covering the four shapes you listed:

```
> The following will run as root:
  · dnf install -y epel-release
      source: uvr's vendored rules and Posit's sysreqs API
      note: adds a third-party repository to this system
  · curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- --no-modify-path -y
      source: uvr's vendored rules
      note: downloads and executes a script from the network
  · dnf install -y --nogpgcheck https://mirrors.rpmfusion.org/free/el/rpmfusion-free-release-$(rpm -E %rhel).noarch.rpm
      source: uvr's vendored rules
      note: installs a package directly from a URL; disables package signature verification
  · sudo dnf install -y gdal-devel geos-devel proj-devel
  Commands without a note are ones uvr matched no pattern in, not ones it has vetted.
```

Three notes on the design:

- **`--nogpgcheck` surfaces on its own terms**, as you asked, and stacks with the URL note rather than replacing it — both facts apply to that command and both are worth printing.
- **Enabling EPEL is classified as a repository addition, not a package install.** `dnf install -y epel-release` is how EPEL gets enabled; calling it a package install would understate what it changes about the machine.
- **The closing line is load-bearing.** `effects` is substring matching over shell strings, deliberately biased toward over-reporting — a spurious note costs a line of reading, a missing one costs the user the disclosure they were owed. That bias makes silence uninformative, so the block says so outright rather than letting the notes read as a whitelist. A test sweeps every command in the snapshot and asserts that anything touching the network carries at least one note, so a future snapshot bump introducing an unrecognised shape fails the suite instead of shipping an unannotated `curl`.

Tests: 511 → 515.

pat-s:
CI note: all 13 red checks are one external cause, not the change.

Every failing job — Windows, macOS arm64, manylinux x86_64/arm64, and the whole distro suite — fails on the same connect timeout:

```
Network error: error sending request for url (https://cran.r-project.org/src/contrib/PACKAGES.gz):
client error (Connect): operation timed out
```

I checked all 13 rather than sampling; each one references that URL. The failing tests are `test_remove_nonexistent_does_not_crash`, `test_update_dry_run_on_empty_project` and `lock_with_binary_capable_source_records_source_urls`, plus the distro suite's `uvr add` — all fetch the CRAN index, none touch sysreqs.

What passed on the same commit rules out a regression: `Test (macos-x86_64, Rosetta)`, `Test (musllinux-x86_64)`, `Test (musllinux-arm64, QEMU)`, `rhel-8`, both build jobs, and the security audit. Same code, different runners.

I can't `gh run rerun --failed` on this repo (contributor, not admin), so a re-run once CRAN is reachable is yours to kick off. Worth watching `rocky-9` in particular — it's the job that exercises the `pre_install` path end to end, and it's the one I'd want green before merge rather than written off as flake.

nbafrank:
Unfortunately not all tests made it through. Please fix and then we can merge!

pat-s:
Tests were fine, they just failed due to a CRAN outage when they first ran. (Did you even look at them and the failure reason? ;)). I even noted that in my previous comment...

nbafrank:
Really sorry for the delay (got really swamped with life, super crazy). And also sorry for not paying attention to the tests. You were totally right, just didn't see it :( it's been merged and looks really great. Keeping track of where the root commands came from was super clever. Thanks for your help :)

Issue #222, opened 2026-08-01
Packages with dots in their name silently install the wrong package when declared as [dependencies.pkg.name]


gladkia:
## Symptom

When a package with a dot in its name is declared using the TOML dotted-key syntax, uvr either silently installs a different package, silently installs nothing, or errors with a confusing message pointing at the wrong name.

```toml
# Intended: install data.table
[dependencies.data.table]
version = "*"
```

```
x ERROR  Package not found: data.
```

## Root cause

TOML parses `[dependencies.X.Y]` as: section `dependencies`, key `X`, sub-key `Y`.
uvr then resolves package `X`, not `X.Y`. Depending on whether `X` is a real CRAN package, three failure modes occur.

### Category 1 — Silent wrong install (most dangerous)

`X` is a real CRAN package → uvr installs `X` and its deps, reports success. The lockfile looks healthy but `X.Y` is absent.

| Intended | Actually installed |
|---|---|
| `shiny.i18n` | `shiny` + 30 deps |
| `shiny.test` | `shiny` + 30 deps |
| `glue.extras` | `glue` (1 pkg) |
| `dplyr.helpers` | `dplyr` + deps |
| `jsonlite.ext` | `jsonlite` (1 pkg) |
| `fake.pkg` | `fake` + MASS + Matrix + Rdpack |

### Category 2 — Silent no-op

`X` is not a CRAN package → uvr resolves 0 packages, reports "0 package(s)", no error.

| Intended | Result |
|---|---|
| `R.utils` | 0 packages (silent) |
| `R.oo` | 0 packages (silent) |
| `R.rsp` | 0 packages (silent) |
| `R.cache` | 0 packages (silent) |

### Category 3 — Visible error (at least detectable)

`X` exists on CRAN but fails to resolve for other reasons, or `X` is not found at all.

| Intended | Error |
|---|---|
| `data.table` | `Package not found: data` |
| `mycomp.utils` | resolution failed |
| `GO.db` | resolution failed |

## Reproduction

```bash
uvr init test --r-version 4.5

# Any of these trigger the bug:
echo '[dependencies.data.table]
version = "*"' >> uvr.toml          # Cat 3: error
echo '[dependencies.shiny.i18n]
version = "*"' >> uvr.toml          # Cat 1: installs shiny
echo '[dependencies.R.utils]
version = "*"' >> uvr.toml          # Cat 2: silent no-op

uvr lock
```

## Workaround

[...]

PR #223, opened 2026-08-01
fix: detect dotted TOML keys in [dependencies] and emit a clear error

gladkia:
## Problem

R package names commonly contain dots (`data.table`, `R.utils`, `shiny.i18n`, `org.Hs.eg.db`). When a user declares these using the TOML table-header syntax, TOML parses the dotted key as a nested structure — silently resolving the wrong package or installing nothing.

**Four failure modes, all silent or misleading:**

| Written | TOML sees | Result |
|---|---|---|
| `[dependencies.shiny.i18n]` | key `shiny`, sub-key `i18n` | installs `shiny` (30 deps) — **wrong package, no warning** |
| `[dependencies.R.utils]` | key `R`, sub-key `utils` | installs 0 packages — **silent no-op** |
| `[dependencies.data.table]` | key `data`, sub-key `table` | `Package not found: data` — confusing |
| `[dependencies.org.Hs.eg.db]` + `bioc = true` | `bioc` attributed to wrong sub-key | resolves `org` from CRAN, not Bioconductor |

Confirmed on uvr 0.4.4, macOS 26 (Intel + Apple M1).

## Fix

After parsing the raw `toml::Value`, validate that every table-valued entry under `[dependencies]` / `[dev-dependencies]` only contains known `DetailedDep` fields (`version`, `bioc`, `git`, `rev`). Any other key signals a dotted-key trap → `ManifestParse` error with an actionable message:

```
dependency `data` in [dependencies] has unexpected sub-key(s): ["table"].

This is usually caused by a dotted TOML key: [dependencies.data.table] is
parsed by TOML as package `data` with sub-key `table`, not as package
`data.table`.

If you meant to declare package `data.table`, use a quoted key:

    [dependencies]
    "data.table" = "*"

Or run: uvr add data.table
```

## Valid usage is unaffected

- `[dependencies.DESeq2]` with `bioc = true` ✅
- `"data.table" = "*"` in flat `[dependencies]` ✅
- `[dependencies."org.Hs.eg.db"]` with `bioc = true` ✅

## Tests

Six new tests in `manifest.rs`:
- `dotted_key_single_dot_is_rejected` — `data.table`
- `dotted_key_r_prefix_is_rejected` — `R.utils`
- `dotted_key_bioc_flag_is_rejected` — `org.Hs.eg.db`
- `dotted_key_dev_dependencies_is_rejected` — `[dev-dependencies.shiny.test]`
- `quoted_dot_package_names_are_accepted` — all four work with quoted keys
- `table_header_syntax_with_valid_fields_is_accepted` — `DESeq2`, `myPkg` still work

Full test suite: 502 passed, 0 failed.

### DIFF
--- a/crates/uvr-core/src/manifest.rs
+++ b/crates/uvr-core/src/manifest.rs
@@ -116,6 +116,10 @@ impl std::str::FromStr for Manifest {
         // `[dependencies.data.table]` creates a nested table under key `data`
         // with a sub-key `table` — not a valid DetailedDep field. This check
         // catches that case before serde silently resolves the wrong package.
+        //
+        // NOTE: `VALID_DEP_KEYS` must list every field of `DetailedDep`. A
+        // field added to that struct without updating this slice will cause
+        // valid manifests to be rejected — keep them in sync.
         const VALID_DEP_KEYS: &[&str] = &["version", "bioc", "git", "rev"];
 
         for section in &["dependencies", "dev-dependencies"] {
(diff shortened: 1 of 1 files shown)


nbafrank:
The bug is real and the detection is right — I reproduced it on main with `[dependencies.data.table]`, which resolves against a package literally named `data` and fails with "Package not found: data.", and confirmed your branch catches it. I also checked for false positives: `[dependencies.DESeq2]{bioc=true}`, `[dependencies.myGitPkg]{git=...,rev=...}`, `[dependencies."org.Hs.eg.db"]{bioc=true}` and flat quoted `"data.table" = "*"` all still parse fine.

**One fix before merge — the remediation message is wrong on your own flagship example.** With `[dependencies.org.Hs.eg.db]` the current output is:

```
dependency `org` in [dependencies] has unexpected sub-key(s): ["Hs"]
...
    "org.Hs" = "*"
Or run: uvr add org.Hs
```

The detection fires correctly, but the advice names `org.Hs` — a package that doesn't exist. The check inspects one nesting level and stops at the first unrecognized key instead of walking the nested-table chain to reconstruct the full dotted name. For Bioconductor annotation packages this is the common case, not an edge case (`org.Hs.eg.db`, `org.Mm.eg.db`, `TxDb.Hsapiens.UCSC.hg38.knownGene` — 2 to 4 dots), and a user following the advice verbatim writes a manifest entry for a nonexistent package. Walk the chain to the leaf and quote the whole thing.

Minor, non-blocking: `VALID_DEP_KEYS` duplicates `DetailedDep`'s field list by hand with no compile-time link — a future field added to the struct will silently start being rejected. Worth a comment on the struct pointing at it.

Thanks for filing #222 alongside — the issue/PR pairing made this quick to verify.

gdevenyi:
Drive-by from #211's neighbourhood — I hit dotted names from the P3M side and wanted to check @nbafrank's remediation report rather than just agree with it. It reproduces, and there is one detail worth having before you fix it.

**The advice is only wrong for names with two or more dots.** I ran your branch against three real cases:

```
[dependencies.data.table]                        advises: uvr add data.table                 ✓ correct
[dependencies.org.Hs.eg.db]                      advises: uvr add org.Hs                     ✗ no such package
[dependencies.TxDb.Hsapiens.UCSC.hg38.knownGene] advises: uvr add TxDb.Hsapiens              ✗ no such package
```

Which explains why it reads as correct at a glance: `data.table` is the flagship example in the PR title and the issue, and it is the one case that happens to work. The check inspects one nesting level, so `key` + `first_unknown` reconstructs the whole name only when there are exactly two segments.

The shape of the TOML is the reason — `[dependencies.org.Hs.eg.db]` nests all the way down:

```
dependencies = { org = { Hs = { eg = { db = { bioc = true } } } } }
```

so `inner.keys()` is `["Hs"]` at the level you inspect, and `eg`/`db` are two and three tables further in. Walking to the leaf gets you the full name, and it also gets you the right *detection* in a case the current code decides by luck: at the leaf, `{bioc = true}` is a table whose keys are all valid `DetailedDep` fields, so the "unknown sub-key" signal only exists at the *upper* levels. Recursing without care would find a valid-looking table at the bottom and conclude nothing is wrong.

Bioconductor annotation packages are where this bites, and they are not exotic — `org.*.eg.db` is one per organism, and the `TxDb.*` / `BSgenome.*` families are four and five segments. A user following the advice verbatim writes a manifest entry for a package that does not exist, which trades a confusing failure for a confident wrong one.

Also worth confirming since it is the thing that makes this safe to fix aggressively: I re-ran @nbafrank's false-positive set on your branch and it holds — `[dependencies.DESeq2]{bioc=true}`, `[dependencies.myGitPkg]{git=...,rev=...}`, `[dependencies."org.Hs.eg.db"]{bioc=true}` (the quoted form, which is the correct spelling and must keep working) and flat `"data.table" = "*"` all still parse.

On the non-blocking `VALID_DEP_KEYS` point: that list is also the only thing standing between a user and this error, so a field added to `DetailedDep` without updating it turns a *valid* manifest into a hard parse error, not just a missed check. Given the constraint is "every field of `DetailedDep`", deriving it beats a comment — `serde`'s field list is available via a `#[derive(Deserialize)]` on a unit-struct visitor, or failing that a `#[test]` that fails when the two lists diverge is cheap and does not need anyone to read a comment at the right moment.

Nice find on the bug itself; I reproduced the original `Package not found: data.` on main before your branch and it is exactly as reported.

gladkia:
@nbafrank , @gdevenyi  - thanks for the thorough review :). The multi-dot reconstruction bug is exactly right, and a bad one given that Bioconductor annotation packages are where you'd hit dotted names most often.

Fixed in the latest commit: instead of stopping at the first unrecognised key, the check now walks the nested table chain to the leaf and joins all keys to reconstruct the full dotted name. `[dependencies.org.Hs.eg.db]` now advises `uvr add org.Hs.eg.db` (not `org.Hs`), and `[dependencies.TxDb.Hsapiens.UCSC.hg38.knownGene]` likewise names the full five-segment package. Two new tests cover these cases.

On the `VALID_DEP_KEYS` / `DetailedDep` sync concern: added a comment on the const pointing at the struct and noting they must stay in sync. A compile-time link would be cleaner (the derive-based visitor approach), but that felt like a bigger change than this PR calls for. Happy to follow up in a separate PR if useful.

nbafrank:
Merged. The chain walk is right, and the two new tests pin exactly the cases that matter — `org.Hs.eg.db` and the five-segment `TxDb.Hsapiens.UCSC.hg38.knownGene` are where users actually meet this.

Thanks also for filing #222 alongside the fix; the issue/PR pairing made the failure mode quick to verify against main. Fixes #222.

PR #224, opened 2026-08-02
Add an installer for Windows

joshuamarie:
Adding `install.ps1`, containing a script for installing `uvr` for Windows. This mirrors the current `install.sh` on the root `main` tree.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -344,6 +344,38 @@ jobs:
       - name: Check every mapped repo still serves
         run: cargo test -p uvr-core --test p3m_repos_live -- --ignored --nocapture
 
+  # install.ps1 has no coverage anywhere above: the windows-x86_64 matrix
+  # entry in `test` builds uvr with cargo directly and never touches the
+  # installer. Run it for real, against the latest published release, the
+  # same way a user invoking `irm ... | iex` would. Catches asset-name
+  # drift, a changed zip layout, and PATH-writing regressions — the three
+  # things gdevenyi flagged as most likely to break silently between
+  # releases (PR #224).
+  test-install-script:
+    name: Test install.ps1 (Windows)
+    runs-on: windows-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Run install.ps1 into a temp dir
+        shell: pwsh
+        run: |
+          $installDir = Join-Path $env:RUNNER_TEMP "uvr-install-smoke"
+          $env:UVR_INSTALL_DIR = $installDir
+          .\install.ps1
+          & (Join-Path $installDir "uvr.exe") --version
+
+      - name: Verify the install dir was added to the user PATH
+        shell: pwsh
+        run: |
+          $installDir = Join-Path $env:RUNNER_TEMP "uvr-install-smoke"
+          $userPath = [Environment]::GetEnvironmentVariable("Path", "User")
+          if ($userPath -notlike "*$installDir*") {
+            Write-Error "install.ps1 did not add $installDir to the user PATH"
+            exit 1
+          }
+          Write-Host "OK: $installDir present in user PATH"
+
   audit:
     name: Security audit
     runs-on: ubuntu-latest
--- a/README.md
+++ b/README.md
@@ -10,9 +10,21 @@ A fast R package and project manager, written in Rust.
 
 `uvr` brings uv-style project management to R: a `uvr.toml` manifest, a reproducible `uvr.lock` lockfile, and a per-project isolated library. Packages install from pre-built [P3M](https://packagemanager.posit.co/) binaries by default — no compilation, no waiting — with automatic fallback to CRAN source. R versions are managed per-project with no `sudo` required.
 
-```sh
-$ curl -fsSL https://raw.githubusercontent.com/nbafrank/uvr/main/install.sh | sh
+1.  Linux / MacOS
 
+    ``` sh
+    curl -fsSL https://raw.githubusercontent.com/nbafrank/uvr/main/install.sh | sh
+    ```
+
+2.  Windows
+
+    ``` sh
+    powershell -ExecutionPolicy Bypass -c "irm https://raw.githubusercontent.com/nbafrank/uvr/main/install.ps1 | iex"
+    ```
+
+Here's a following demo of `uvr`: 
+
+``` sh
 $ uvr init my-analysis
 $ uvr add ggplot2 dplyr tidymodels
 $ uvr sync          # installs from lockfile, idempotent
--- a/README.md
+++ b/README.md
@@ -18,8 +18,8 @@ A fast R package and project manager, written in Rust.
 
 2.  Windows
 
-    ``` sh
-    powershell -ExecutionPolicy Bypass -c "irm https://raw.githubusercontent.com/nbafrank/uvr/main/install.ps1 | iex"
+    ``` powershell
+    irm https://raw.githubusercontent.com/nbafrank/uvr/main/install.ps1 | iex
     ```
 
 Here's a following demo of `uvr`: 
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ R has several package management tools — `renv`, `pak`, `rv`, `rig` — each s
 
 Here is how existing tools compare and where the gaps are:
 
-- **renv** — the de-facto standard for reproducibility. It snapshots an existing library into a lockfile, but it does not manage R versions ("renv tracks, but doesn't help with, the version of R used") and relies on `install.packages()` under the hood, which is slow and requires compilation on Linux.
+- **renv** — the de-facto standard for reproducibility. It snapshots an existing library into a lockfile, but it does not pin R versions ("renv tracks, but doesn't help with, the version of R used") and relies on `install.packages()` under the hood, which is slow and requires compilation on Linux.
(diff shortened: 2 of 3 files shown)


gdevenyi:
Nice to see the Windows gap closed — the install path being POSIX-only has been the awkward part of "uvr manages R for you" on Windows. I reviewed it from the packaging side since I have been living in `install.sh`'s assumptions for #211. Three things, one of which I would fix before merge.

**Verified working first**, so the rest reads in context. The asset naming and archive layout are right — I checked against the real v0.4.4 release rather than the code:

```
uvr-x86_64-pc-windows-msvc.zip   ->  uvr.exe   (single entry, at root)
sha256sums.txt line 5            ->  c56c2100…fcac  uvr-x86_64-pc-windows-msvc.zip
```

and the checksum logic verifies correctly against it under pwsh:

```
expected: c56c21002158e741822f2137d23acc08ef349de60d1a787afd47bfc9e7a1fcac
actual:   c56c21002158e741822f2137d23acc08ef349de60d1a787afd47bfc9e7a1fcac
match:    True
```

The `Write-Log`-inside-`Get-Target` pattern is also correct, which is easy to get wrong: `Write-Host` bypasses the pipeline, so the ARM64 branch's log line does not corrupt the returned target string. If that were `Write-Output` the function would return an array and the download URL would be garbage.

---

**1. The PATH write can permanently damage the user's PATH.** This is the one I would fix before merge.

```powershell
$userPath = [Environment]::GetEnvironmentVariable("Path", "User")
$newPath  = if ($userPath) { "$userPath;$InstallDir" } else { $InstallDir }
[Environment]::SetEnvironmentVariable("Path", $newPath, "User")
```

`HKCU\Environment\Path` is very often `REG_EXPAND_SZ` containing literal `%USERPROFILE%`-style references. `GetEnvironmentVariable(..., "User")` returns that **already expanded**, and `SetEnvironmentVariable(..., "User")` writes back a plain `REG_SZ`. So the round-trip silently rewrites every `%VAR%` entry in the user's PATH to its value at install time and drops the expansion type — which breaks on a roaming profile, a changed username, or anything else that made those entries variables in the first place. It is not reversible from the script's side, and the user has no idea it happened.

The usual fix is to go to the registry directly and preserve the kind:

```powershell
$key = "HKCU:\Environment"
$raw = (Get-Item $key).GetValue("Path", "", "DoNotExpandEnvironmentNames")
$kind = (Get-Item $key).GetValueKind("Path")
Set-ItemProperty -Path $key -Name Path -Value "$raw;$InstallDir" -Type $kind
```

I could not run this half — no Windows box — so please confirm on a real machine, but the `GetEnvironmentVariable` expansion behaviour is documented .NET and this is a well-trodden installer footgun.

Related and much smaller: `Test-InPath` compares with `-contains`, which is exact-match. A user whose PATH holds `…\.local\bin\` with a trailing separator, or differing case, gets a duplicate appended on every re-run. `-icontains` plus a `TrimEnd(\)` on both sides handles it.

**2. Checksum verification fails open, silently.** `Get-ExpectedHash` returns `$null` both when `sha256sums.txt` 404s (warned) and when the asset's line is simply absent (not warned), and the caller then skips verification entirely:

```
-- missing asset line (simulates a stripped/tampered sums file) --
returns:  <null>
guard: SKIP VERIFICATION SILENTLY
```

To be fair to this PR: **`install.sh` has the identical hole** (`if [ -n "$EXPECTED" ]`), so this is parity, not a regression, and I would not block on it. But a downloaded-over-the-network checksum file whose absence disables checking is not really a check, and this PR is the moment someone is looking at that code. If it is worth fixing it is worth fixing in both — happy to send that as its own PR against `install.sh` so it does not enlarge this one.

**3. Nothing runs it.** `ci.yml` already has a `windows-latest` runner. A step that dot-sources the script with `UVR_INSTALL_DIR` pointed at a temp dir and asserts `uvr.exe --version` would catch asset-name drift, a zip-layout change, and the PATH logic — the three things most likely to break silently between releases, and none of them visible until a user reports it.

Two smaller notes:

- The README suggests `powershell -ExecutionPolicy Bypass -c "irm … | iex"`. `irm | iex` never touches the execution policy — it is not running a script *file* — so the flag is unnecessary and teaches a habit worth not teaching. `irm https://… | iex` on its own is enough.
- ARM64 → x64-under-emulation is the right call given no `aarch64-pc-windows-msvc` asset exists today, and I like that it says so rather than failing. Worth a `# TODO` pointing at whichever issue tracks a native ARM64 Windows build, so the branch gets revisited rather than becoming permanent.

joshuamarie:
Alright, I tried some fixing for 1 via the registry route you outlined. Upon reading, I confirmed it preserves the `REG_EXPAND_SZ` type on a round-trip. Also folded in the trailing-backslash/case dedup.

nbafrank:
Welcome, and thanks — this is a genuinely careful first contribution. Things I want to call out as *right*, because they're the parts most installer scripts get wrong: forcing TLS 1.2 up front; reading the user PATH with `DoNotExpandEnvironmentNames` and writing it back with the original value kind preserved (so an `REG_EXPAND_SZ` PATH with `%USERPROFILE%` entries survives — most scripts flatten it and quietly break other tools); the case-insensitive trailing-slash-tolerant dedupe before appending; cleanup in `finally`; and updating the session PATH so the closing `uvr --version` smoke test actually runs. The ARM64→x64-under-emulation fallback with a logged TODO is the right call today.

One fix before merge, two nits:

**Fix — checksum parity with install.sh.** When `sha256sums.txt` can't be fetched, this script warns and proceeds. `install.sh` runs under `set -e`, so its failed sums download aborts the install — the sums file ships with every release, so its absence means something is wrong (truncated release, tampering, network middlebox), and that's exactly when you don't want to install the binary anyway. Make the fetch failure a `Write-ErrExit` instead of a warning, and treat "asset not found in the sums file" the same way.

**Nit 1 — `UVR_VERSION` v-prefix.** Tags are `v0.4.5`; a user who sets `UVR_VERSION=0.4.5` gets a 404 with no hint. One line: prepend `v` when it's missing. (install.sh has the identical gap — if you fix it here I'll mirror it there, or feel free to touch both.)

**Nit 2 —** `Set-ItemProperty` on HKCU\Environment doesn't broadcast `WM_SETTINGCHANGE`, so already-running Explorer won't propagate the new PATH to terminals it spawns until re-login. Your "open a new terminal" message covers the common case; a comment noting the limitation would spare the next person a debugging session.

With the checksum fix this merges. The README additions look good and match the sh installer's docs.

joshuamarie:
Alright, thanks for the bother. Now, going through each point:

1. *PATH write* already uses the registry-safe approach you outlined (`DoNotExpandEnvironmentNames`, preserving the value kind, `-ieq` + `TrimEnd("\")` on both sides for the dedup check), so this should be in good shape. I don't have a Windows box either, but I've added a CI job (`test-install-script`) that runs `install.ps1` end-to-end on `windows-latest` and asserts the install dir lands in the registry `Path` — not a substitute for testing on a real roaming profile, but it's real signal on every push now instead of none.

2. On *Checksum fails open*, this is indeed a parity with `install.sh`, not a regression, so I left the actual skip-on-missing behavior alone. Added a warning for the case where `sha256sums.txt` exists but has no line for the asset (it only warned on the file-missing case before) so the silent path at least shows up in the log. Happy to have that superseded by whatever you send against `install.sh`.

3. Then, added `test-install-script` to the current `ci.yml`: runs the real installer against the latest release, checks `uvr.exe --version`, and checks the PATH write. Windows CI now actually exercises this path.

On `README`, I then removed `-ExecutionPolicy Bypass`, complying you and @gdevenyi. 

Let me know if the CI job needs adjusting. I didn't want to gate it on a specific tag since it hits the live release like a real user would.

nbafrank:
@joshuamarie — checking in. This is genuinely good work and I'd like to land it; it's the last thing standing between Windows users and a one-line install.

There's only one required change: make a missing or unfetchable `sha256sums.txt` abort the install rather than warn and continue. `install.sh` runs under `set -e`, so its failed sums download already aborts — the sums file ships with every release, so its absence means something is wrong (truncated release, tampering, a middlebox), which is exactly when you don't want to install the binary anyway.

The two nits are optional: prepending `v` when `UVR_VERSION` is set without it, and a comment noting that `HKCU\Environment` changes don't broadcast `WM_SETTINGCHANGE`.

If you've moved on or would rather I finish it, just say — I'm happy to make the checksum change myself and merge with you credited. No pressure either way, and thanks for the contribution; the registry handling in particular was more careful than most installers manage.

joshuamarie:
Alright, done. I pushed the latest abort-on-missing-checksum patches plus both nits. Appreciate the offer, but this one was small enough to finish myself. Thanks for the thorough review, especially catching the `REG_EXPAND_SZ` issue earlier. This would've been a nasty one to debug for whoever hit it first.

nbafrank:
Merged — thanks for seeing it through, and for a genuinely careful first contribution.

The `REG_EXPAND_SZ` handling is the part I'd highlight to anyone reading this later: reading the user PATH with `DoNotExpandEnvironmentNames` and writing it back with the original value kind means a PATH containing `%USERPROFILE%` survives intact. Most installer scripts flatten it and quietly break other tools, and nobody notices until something else fails weeks later.

Windows users can now install with a single line, which closes the last platform gap in getting started.

PR #229, opened 2026-08-03
fix: prepend R lib dir to LD_LIBRARY_PATH instead of replacing it

### LINKED ISSUE #228: R CMD INSTALL subprocess clobbers LD_LIBRARY_PATH, breaking source installs on HPC clusters with module-provided BLAS/LAPACK
## Symptom

On HPC clusters where BLAS/LAPACK are provided by the environment module system (e.g. OpenBLAS, MKL, FlexiBLAS loaded via `module load`), source package installs fail during lazy-loading with:

```
libRlapack.so: cannot open shared object file: No such file or directory
```

or:

```
unable to load shared object '.../Matrix/libs/Matrix.so':
  libRblas.so => not found
  libRlapack.so => not found
  libR.so => not found
```

## Root cause

`r_cmd_install.rs` sets `LD_LIBRARY_PATH` to R's own lib directory:

```rust
cmd.env("LD_LIBRARY_PATH", r_lib_str.as_ref())
```

This *replaces* the entire `LD_LIBRARY_PATH` inherited from the parent process. On HPC systems, that value is set by the module system to expose BLAS/LAPACK shared libraries (`libopenblas.so`, `libmkl_rt.so`, etc.). After uvr clobbers it, those libraries become invisible to the R CMD INSTALL subprocess during its lazy-loading phase.

## Reproduction

A typical HPC setup:

```bash
module load R/4.5    # R built with OpenBLAS, sets LD_LIBRARY_PATH to include OpenBLAS libs
uvr add --bioc SummarizedExperiment   # → S4Arrays fails
# ERROR: lazy loading failed for package 'S4Arrays'
# libRlapack.so: cannot open shared object file
```

P3M binaries for packages like `Matrix` link against `libRblas.so` and `libRlapack.so`. When R is built against OpenBLAS without bundling those as separate files, the libraries exist only via the module-provided `LD_LIBRARY_PATH`. Setting symlinks (`libRblas.so → libopenblas.so`) in a user-visible path and exporting `LD_LIBRARY_PATH` from the parent shell has no effect — uvr resets the variable before spawning R CMD INSTALL, so the subprocess never sees them.

Tested on a HPC cluster (SLURM, RHEL-compatible, GLIBC 2.28, OpenBLAS as BLAS provider).

## Fix

Prepend R's lib dir instead of replacing the existing value — the same pattern already used for `PATH` in the same function:

```rust
// PATH (existing, correct pattern):
let new_path = match std::env::var("PATH") {
    Ok(p) if !p.is_empty() => format!("{bin}:{p}"),
    _ => bin,
};

// LD_LIBRARY_PATH (this fix — same pattern):
let prepend_lib = |var: &str| -> String {
    match std::env::var(var) {
        Ok(existing) if !existing.is_empty() => format!("{r_lib_str}:{existing}"),
        _ => r_lib_str.to_string(),
    }
};
cmd.env("LD_LIBRARY_PATH", prepend_lib("LD_LIBRARY_PATH"))
```

See PR #229 .

## Affected environments

Any system where:
- BLAS/LAPACK are separate shared libraries loaded via environment modules
- R is built `--with-blas=openblas` or similar (no bundled `libRlapack.so`)
- Source packages use lazy-loading (Bioconductor core: `S4Arrays`, `SummarizedExperiment`, `Matrix`, `BumpyMatrix`, `MultiAssayExperiment`)

This covers standard HPC setups: SLURM clusters, LSF clusters, university HPC, enterprise sHPC environments.

## Packages that work without the fix

CRAN binary packages without R-internal BLAS/LAPACK dependencies (`data.table`, `ggplot2`, etc.) install correctly — confirming the regression is specific to packages that link against `libRblas.so`/`libRlapack.so`.

gladkia:
## Problem

`R CMD INSTALL` subprocesses spawned by `uvr` fail on HPC clusters where BLAS/LAPACK are provided by an environment module system (OpenBLAS, MKL, FlexiBLAS). The failure manifests during lazy-loading of packages that link against `libRlapack.so` or `libRblas.so`:

```
libRlapack.so: cannot open shared object file: No such file or directory
```

The root cause: `build_cmd` set `LD_LIBRARY_PATH` to R's own lib dir, *replacing* the value inherited from the parent process. The module system had set that value to expose BLAS/LAPACK — after uvr clobbered it, those libraries became invisible to the install subprocess.

A second failure mode: even with `LD_LIBRARY_PATH` correct for the top-level process, packages that link against `libRblas.so`/`libRlapack.so` can fail when R spawns its own child process for lazy loading — that child does not go through `build_cmd`, so it never inherits the corrected `LD_LIBRARY_PATH`.

Affected packages on typical Bioconductor projects: `S4Arrays`, `SummarizedExperiment`, `BumpyMatrix`, `MultiAssayExperiment`, `Matrix`. Standard HPC setups (SLURM, LSF, university clusters) are the common case — tested on a SLURM cluster with RHEL-compatible OS, GLIBC 2.28, and a custom-compiled R with OpenBLAS.

Closes #228.

## Root cause detail

There are two related but distinct paths where the library path was wrong:

1. **`r_cmd_install.rs` / `build_cmd`** — `LD_LIBRARY_PATH` replaced instead of prepended. Affects all packages with compiled code linking against `libRblas.so`/`libRlapack.so`.

2. **`r_env.rs` / `REnv::vars()`** — Same clobber for `uvr run` and `uvr activate`. A user who activates a project environment and then loads a package that needs BLAS gets the same failure at runtime, not just at install time.

## Fix

Three changes, in order:

### 1. Prepend in `build_cmd` (`r_cmd_install.rs`)

```rust
let prepend_lib = |var: &str| -> String {
    match std::env::var(var) {
        Ok(existing) if !existing.is_empty() => format!("{r_lib_str}:{existing}"),
        _ => r_lib_str.to_string(),
    }
};
cmd.env("LD_LIBRARY_PATH", prepend_lib("LD_LIBRARY_PATH"))
```

Mirrors exactly how `PATH` is extended in the same function.

### 2. Prepend in `REnv::vars()` (`r_env.rs`)

Same pattern applied to `uvr run` and `uvr activate`, so the runtime environment is consistent with what `uvr sync` produces.

### 3. `R_LD_LIBRARY_PATH` — guarded by `libR.so` presence

R's wrapper script (`bin/R`) sources `etc/ldpaths`, which sets `R_LD_LIBRARY_PATH = ${R_HOME}/lib` **only when the variable is not already set**. R then prepends `R_LD_LIBRARY_PATH` to `LD_LIBRARY_PATH` in every subprocess it spawns (byte-compilation children, `Rscript` calls in build scripts, etc.). Setting it ensures that BLAS/LAPACK paths propagate transitively through R's entire process tree, not just the top-level `R CMD INSTALL`.

**Guard:** On Posit portable builds, `libR.so` lives one level deeper (`lib/R/lib/`) than `r_lib_dir` (`lib/`). Pre-setting `R_LD_LIBRARY_PATH` to the wrong path suppresses `ldpaths`'s correct default and causes `exec/R: error while loading shared libraries: libR.so`. We only set `R_LD_LIBRARY_PATH` when `libR.so`/`libR.dylib` is actually present in `r_lib_dir`:

```rust
let libr_present = r_lib_dir.join("libR.so").exists()
    || r_lib_dir.join("libR.dylib").exists();
if libr_present {
    cmd.env("R_LD_LIBRARY_PATH", prepend_lib("R_LD_LIBRARY_PATH"));
}
```

On an HPC cluster where R is compiled from source, `libR.so` is directly in `r_lib_dir` — guard passes, BLAS paths propagate through all of R's own children. On Posit portable builds, the guard is false, `ldpaths` remains free to set the correct `R_HOME/lib`, and `exec/R` starts normally.

## Testing

- All 585 tests pass across all platforms (macOS arm64/x86_64, Windows, manylinux x86_64/arm64, musllinux x86_64/arm64, distro matrix).
- Root cause confirmed on HPC cluster (SLURM, RHEL-compatible, GLIBC 2.28, custom-compiled R + OpenBLAS). With the fix applied: `BiocGenerics`, `SummarizedExperiment`, `BumpyMatrix`, `data.table` all install correctly.
- `R_LD_LIBRARY_PATH` guard verified: manylinux CI smoke test (`uvr run check.R`) passes, confirming Posit portable builds are unaffected.
- No change to Windows behaviour (`LD_LIBRARY_PATH` branch is a runtime `if cfg!(target_os = "windows")` — compiles on all platforms).


gdevenyi:
Reviewed this because it neighbours the distro-suite work in #232/#233 — the reporting environment (RHEL-compatible, glibc 2.28) is one the matrix covers. **The fix is correct and the pattern is right, but it repairs one of three call sites**, and I think that matters for whether #228 actually closes.

## The empty-string guard deserves calling out first

```rust
Ok(existing) if !existing.is_empty() => format!("{r_lib_str}:{existing}"),
```

That guard is the best detail in the diff and it is the part a reviewer skims past. Without it, an empty-but-set `LD_LIBRARY_PATH` yields `<r_lib>:` — and a trailing empty entry is interpreted by glibc's loader as *the current directory*. A naive `unwrap_or_default()` version would have quietly introduced a code-execution footgun into a subprocess that compiles code. Getting this right without being asked is the reason I would like it pinned by a test (below).

Prepending is otherwise clearly right: R's own lib dir keeps priority for `libR.so`, which is why the code exists, while inherited entries stay reachable.

## Main issue: the sibling call site is untouched

`crates/uvr-core/src/r_env.rs:105-106` has the identical bug:

```rust
("DYLD_LIBRARY_PATH", r_lib_dir.clone()),
("LD_LIBRARY_PATH", r_lib_dir),
```

and that function documents itself as the authority:

> **This is the single source of truth**: `uvr run` applies these to the R child process, `uvr activate` exports the same pairs into the shell.

| path | site | fixed by this PR |
|---|---|---|
| `uvr sync` / `uvr add` -> `R CMD INSTALL` | `r_cmd_install.rs:397` | yes |
| `uvr run script.R` | `r_env.rs:105` via `run.rs:80` | **no** |
| `uvr activate` (sourced) | `r_env.rs:105` via `activate.rs` | **no** |

So on the cluster in #228, `uvr sync` will now succeed and then `uvr run` will fail with the same `libRlapack.so: cannot open shared object file` — same cause, one layer over. That is a worse outcome to debug than the current consistent failure, because the fix appears to have worked.

The activation case is the sharpest of the three: it replaces `LD_LIBRARY_PATH` in the user's **interactive shell**, so a module environment is broken for everything in that session, not only R. It is at least reversible — `managed_vars` saves and restores it — but during activation the module paths are gone.

Two things encode the old behaviour and would need updating alongside it: the comment at `activate.rs:53-56` ("`PATH` is handled specially (it is prepended to, not replaced); the rest are set to literal values"), and the test at `r_env.rs:194-195`, which asserts exact equality:

```rust
assert_eq!(get("LD_LIBRARY_PATH"), lib);
```

## Worth knowing: prepend may not deliver what the description promises

I checked the portable build rather than assuming. From `lib/R/etc/Makeconf`:

```
BLAS_LIBS   = -L"$(R_HOME)/lib$(R_ARCH)" -lRblas
LAPACK_LIBS =
```

and the tarball ships `lib/R/lib/libRblas.so` but **no `libRlapack.so` at all** — LAPACK is `lib/R/modules/lapack.so`.

With R's lib dir prepended, `libRblas.so` therefore resolves to R's own reference BLAS and shadows your `libRblas.so -> libopenblas.so` symlink; only `libRlapack.so` falls through to your directory, because that is the one place it exists. It will load, but the result is R's reference BLAS paired with OpenBLAS-as-LAPACK rather than "the cluster's OpenBLAS".

Not a blocker, and prepend is still strictly better than replace. But appending is what would actually hand you OpenBLAS for both — at the cost of the `libR.so` resolution this code exists to guarantee, so I do not think you should. I would just retitle the claim: this **unbreaks module-provided libraries** rather than making uvr use module BLAS. Worth being precise about, because someone will later measure BLAS performance and be surprised.

It also suggests a separate question, not for this PR: uvr's portable R provides no `libRlapack.so`, so any binary package built against a LAPACK-linking R needs a symlink workaround like yours. That may deserve its own issue.

## Test coverage

None here, and it is cheap: `build_cmd` returns a `Command` and `Command::get_envs()` is stable. Three cases — unset, empty, populated — would pin the behaviour, and the empty case especially, since that is the one most likely to be "simplified" away by someone who reads the guard as redundant.

## Minor

- The description says the branch is `#[cfg(not(windows))]`. It is `if cfg!(target_os = "windows") { ... } else { ... }` — a runtime branch that still compiles on Windows. Same behaviour, but the distinction matters if anyone later adds platform-specific types there.
- `DYLD_LIBRARY_PATH` is set on Linux too (pre-existing, inert). Also note macOS SIP strips `DYLD_*` when spawning protected binaries, so inheriting it there may be moot in practice — again pre-existing, not this PR's problem.

## Security

Prepending means uvr inherits a caller-controlled `LD_LIBRARY_PATH` into a subprocess, which sounds worse than it is: `R CMD INSTALL` already compiles and runs arbitrary package code, so the threat model does not move; R's own directory takes precedence; and the empty-entry case is guarded. No objection.

## Recommendation

Approve the direction. I would fold `r_env.rs` into this PR rather than a follow-up — it is the same three-line change, and shipping only half means #228 stays open in practice while looking closed. The unit test would be worth it too, mostly to protect the empty-string guard.

Good diagnosis, and the `dyn.load` isolation experiment in the description is exactly the right way to have proved the root cause without a patched binary.

nbafrank:
Good find, and the diagnosis is right — I traced it and the prepend logic you wrote is well-built (the guard against an empty inherited value correctly avoids a stray colon, Windows is untouched, and putting the managed R's lib dir first preserves isolation for anything it does provide while letting the module system's BLAS fall through, which is exactly the behaviour #228 needs).

**But it fixes one of the three places uvr does this, and the other two hit the same users one command later.**

Your branch predates #197 (shell activation), which landed a shared environment builder — `crates/uvr-core/src/r_env.rs:99-110`, `REnv::vars()`:

```rust
("DYLD_LIBRARY_PATH", r_lib_dir.clone()),
("LD_LIBRARY_PATH", r_lib_dir),
```

That unconditionally replaces, never reads the inherited value, and it's the single source of truth for **`uvr run`** (`run.rs:79-82`) and **`uvr activate`** (`activate.rs:208`, all three shell emitters). So after this merges, `uvr add --bioc SummarizedExperiment` works on your cluster — but `uvr run -e 'library(SummarizedExperiment)'`, or `source .uvr/activate` followed by anything, still clobbers the module system's OpenBLAS/MKL and dies with the same `libRlapack.so` error. That's the normal workflow right after a successful sync, so it'll be hit immediately.

**Asks:**

1. Apply the same prepend in `REnv::vars()` so `run` and `activate` behave like `R CMD INSTALL` — or, if you'd rather keep this PR tight, retitle it to "R CMD INSTALL only" and file a follow-up. I'd rather not merge a silent inconsistency, since #228 reads as general HPC breakage rather than install-only.
2. Add a unit test for the env construction — `build_cmd` is reachable from the module's own test mod and `Command::get_envs()` lets you assert without spawning, and the file already has the `env_lock()` pattern for safe env mutation. Pin both cases: inherited value present → prepended with `:`, absent/empty → no stray colon.
3. Cheap insurance: if `r_lib_str` itself is ever empty (a malformed `r_binary` with no parent), the result is `":$LD_LIBRARY_PATH"` — a leading empty component means "current directory" to the linker. One-line guard.

Also worth rebasing — you're ~51 commits behind main now, and `r_env.rs` didn't exist when you branched.

gladkia:
Good points — updated in a second commit on the same branch (needs a force-push to land, local branch is rebased on main).

**`REnv::vars()` fixed:** applied the same prepend logic to lines 105-106 in `r_env.rs`. `uvr run` and `uvr activate` now preserve the inherited `LD_LIBRARY_PATH` just like `R CMD INSTALL` does after the first commit. The comment in `vars()` explains why.

**Three new unit tests in `r_env.rs`:**
- `vars_prepend_r_lib_dir_to_inherited_ld_library_path` — module BLAS path (`/apps/rocs/OpenBLAS/lib`) survives into the output; asserts prepend-with-separator, not replace
- `vars_empty_r_lib_dir_does_not_produce_leading_separator` — empty `r_lib_dir` (malformed binary path) keeps the inherited value intact, no leading `:`
- existing `vars_isolate_the_project_from_system_libraries` updated to clear inherited `LD_LIBRARY_PATH` so it is deterministic under any test runner

All three use the existing `env_lock()` pattern from the file for safe env mutation. `Command::get_envs()` isn't used here since `vars()` returns a `Vec` directly — asserting on that is simpler and doesn't need a real `Command`.

**Guard:** the empty-`r_lib_dir` branch in `prepend_lib` falls through to the inherited value unchanged rather than producing `":existing"`. One-liner.

584/584 tests pass on the rebased branch.

gladkia:
Two more commits landed since the last update — here is a summary of what changed and why.

**What the previous two commits added** (already discussed above):
- `r_env.rs` prepend — same prepend logic applied to `REnv::vars()` so `uvr run` and `uvr activate` behave like `R CMD INSTALL`.
- Unit tests: `vars_prepend_r_lib_dir_to_inherited_ld_library_path`, `vars_empty_r_lib_dir_does_not_produce_leading_separator`, and the existing isolation test updated to clear the inherited value for determinism under any test runner. All use the existing `env_lock()` pattern.

**What the two new commits add (R_LD_LIBRARY_PATH):**

While testing on an HPC cluster, a second failure class surfaced: packages that link against `libRblas.so`/`libRlapack.so` can still fail during byte-compilation even after the `LD_LIBRARY_PATH` prepend is in place. The error is the same `libRlapack.so: cannot open shared object file` but from a subprocess that R itself spawns for lazy loading — not from `build_cmd` — so the `LD_LIBRARY_PATH` that uvr sets on the top-level `R CMD INSTALL` process is not inherited by that grandchild.

R has a dedicated variable for exactly this: `R_LD_LIBRARY_PATH`. The R wrapper script sources `etc/ldpaths`, which prepends `R_LD_LIBRARY_PATH` to `LD_LIBRARY_PATH` in **every** subprocess R spawns — byte-compilation children, `Rscript` calls in build scripts, etc. Setting it means BLAS paths propagate transitively through the entire R process tree.

The complication: `ldpaths` sets `R_LD_LIBRARY_PATH = ${R_HOME}/lib` **only when the variable is not already set**. On Posit portable builds, `r_lib_dir()` computes `<install_root>/lib`, but `libR.so` lives in `<install_root>/lib/R/lib/` — one level deeper. Pre-setting `R_LD_LIBRARY_PATH` to the wrong path suppresses `ldpaths`'s correct default and causes `exec/R: error while loading shared libraries: libR.so` — exactly the regression we saw in the first attempt.

**Guard:** only set `R_LD_LIBRARY_PATH` when `libR.so`/`libR.dylib` is present in `r_lib_dir`. On an HPC cluster with source-compiled R, `libR.so` is directly in `r_lib_dir` — guard passes, BLAS paths propagate. On Posit portable builds the guard is false, `ldpaths` sets the correct path, `exec/R` starts normally.

**Result: 14/14 CI jobs pass** (manylinux x86_64 + arm64, musllinux, macOS arm64 + x86_64 Rosetta, Windows, full distro matrix). Total: 585 tests.

nbafrank:
Merged — and this ended up better than what the review asked for. Extending the prepend to `REnv::vars()` was the required fix, but the `R_LD_LIBRARY_PATH` propagation is the part I want to call out: catching that R's own subprocesses lose the module paths, *and* that pre-setting it unconditionally would suppress the portable builds' `ldpaths` default and break `exec/R` at startup — that guard on `libR.so` actually being present is exactly the kind of interaction most fixes miss. The empty-lib-dir guard against a leading separator is in as asked, with tests pinning both directions.

585 tests green on the merge. #228 closes with this. Thanks for the thorough follow-through — this will matter to every module-system cluster running uvr.

PR #232, opened 2026-08-03
ci: derive the portable-R glibc floor from the host, not from each matrix entry

gdevenyi:
**Stack position: merge first.** Fixes 4 of the 5 failures in #230; the alpine-320 half is a separate PR that builds on this one.

Closes 4/5 of #230.

## What broke

The first nightly after #211's pr lane went green failed on `rocky-8`, `almalinux-8`, `oracle-8` and `opensuse-155` — all at `uvr r install`, all with the same message:

```
```

That is uvr behaving correctly. The suite was wrong.

## Why

#211 turned "expected to fail at `uvr r install`" from prose in each entry's `note` into an asserted `expect` block. It did that by **matching the prose** — and the prose had only ever been written for three entries. Every sub-floor distro whose note nobody had gotten around to writing sailed through unchanged.

I captured `ldd --version` from all 29 images rather than reasoning about it. Seven entries are below the floor; three had the block:

| entry | glibc | had `expect` |
|---|---|---|
| rhel-8 | 2.28 | yes |
| ubuntu-2004 | 2.31 | yes |
| debian-11 | 2.31 | yes |
| **rocky-8** | 2.28 | **no** |
| **almalinux-8** | 2.28 | **no** |
| **oracle-8** | 2.28 | **no** |
| **opensuse-155** | 2.31 | **no** |

`rhel-8` was the *only* sub-floor entry in the `pr` lane, which is exactly why the pr lane looked complete and the nightly did not.

## The fix

Stop recording it per entry. The floor is a property of the host, so ask the host — compare `ldd --version` against `R_GLIBC_FLOOR`, which mirrors `MANYLINUX_GLIBC_MIN` in `downloader.rs`. The three hand-written `expect` blocks are removed.

A per-entry expectation is only ever as complete as whoever remembered to write it. Deriving it means a distro added later is covered without anyone realising they had to do anything — and it gains an assertion the per-entry version could not express: uvr installing R **below** the floor now fails too.

The floor lands exactly on the RHEL 8 → 9 line. Every `*-9` entry reports precisely `2.34`, so `>= floor` is load-bearing rather than decorative, and the comparison is unit-tested across `2.28 … 3.0` including that boundary.

## Verified

Every reachable sub-floor image, plus the boundary, with a binary built by `ci/build-linux-gnu.sh`:

```
rocky-8       exit=0  expected failure confirmed: glibc 2.28 < 2.34
almalinux-8   exit=0  expected failure confirmed: glibc 2.28 < 2.34
oracle-8      exit=0  expected failure confirmed: glibc 2.28 < 2.34
opensuse-155  exit=0  expected failure confirmed: glibc 2.31 < 2.34
ubuntu-2004   exit=0  expected failure confirmed: glibc 2.31 < 2.34
debian-11     exit=0  expected failure confirmed: glibc 2.31 < 2.34
rocky-9       exit=0  binary-smoke-ok  sysreqs-smoke-ok   <- exactly 2.34, normal path
ubuntu-2204   exit=0  binary-smoke-ok  sysreqs-smoke-ok
```

## One thing worth naming

This is the **third** glibc number in the tree, and they are independent — it has now confused more than one reader:

| number | gates | set in |
|---|---|---|
| 2.34 | whether the portable R builds run at all | `MANYLINUX_GLIBC_MIN`, downloader.rs |
| 2.28 | whether P3M binary *packages* come from the portable repo | `PPM_MANYLINUX_GLIBC_MIN`, downloader.rs |
| 2.28 | whether the uvr binary itself starts | `GLIBC_FLOOR`, ci/build-linux-gnu.sh (#212) |

#212 is why `uvr --version` and `uvr doctor` run on RHEL 8 at all — visible in the failing logs, where the binary starts fine and only `r install` refuses. It cannot move the first row: that is Posit's build, not uvr's.

## Note on `expect`

The mechanism is left in place but has no user after this change. Its next one is the alpine-320 PR, where the condition is genuinely per-entry rather than a host property. If that PR is rejected, `expect` should be deleted rather than kept on spec.

https://claude.ai/code/session_01B2dAmERdjdAJr7xotiKQPB

### DIFF
--- a/ci/distro-suite.sh
+++ b/ci/distro-suite.sh
@@ -225,6 +225,38 @@ UVR="$DIST/uvr-$TARGET/uvr"
 note "install.sh would fetch: uvr-$TARGET.tar.gz"
 [ -x "$UVR" ] || fail "no release build for $TARGET under $DIST"
 
+# Whether the portable R builds can run here at all.
+#
+# Posit publishes them as manylinux_2_34, so a glibc host below 2.34 gets no R
+# — `MANYLINUX_GLIBC_MIN` in r_version/downloader.rs, and a different number
+# from both the 2.28 floor the uvr *binary* is built against (#212) and the
+# 2.28 `PPM_MANYLINUX_GLIBC_MIN` that gates binary *packages*. Three floors,
+# three meanings; only this one decides whether `uvr r install` can work.
+#
+# Derived from the host, deliberately. This used to be an `expect` block on
+# each entry, which meant it was only as complete as whoever remembered to add
+# one — and it was not: rocky-8, almalinux-8, oracle-8 and opensuse-155 are all
+# below the floor, none of them carried the block, and all four failed the
+# first nightly after the pr lane went green (#230). The floor is a property of
+# the host, so the host is what gets asked.
+R_GLIBC_FLOOR=2.34
+# Set by below_r_glibc_floor; reported by the stage below.
+hv=""
+
+below_r_glibc_floor() {
+    # musllinux builds carry no glibc floor; Alpine is judged on its own terms.
+    if [ "$libc" = musl ]; then return 1; fi
+    hv="$(ldd --version 2>&1 | grep -oE '[0-9]+\.[0-9]+' | head -1)"
+    # An unreadable ldd must not silently excuse a distro. Expect success and
+    # let the stage fail loudly if that was wrong.
+    if [ -z "$hv" ]; then return 1; fi
+    hmaj="${hv%%.*}"; hmin="${hv#*.}"; hmin="${hmin%%.*}"
+    fmaj="${R_GLIBC_FLOOR%%.*}"; fmin="${R_GLIBC_FLOOR#*.}"
+    if [ "$hmaj" -lt "$fmaj" ]; then return 0; fi
+    if [ "$hmaj" -eq "$fmaj" ] && [ "$hmin" -lt "$fmin" ]; then return 0; fi
+    return 1
+}
+
 # `--version` on a bare image is the whole glibc-floor question in one line: a
 # gnu binary linked against a newer glibc than this distro ships dies right
 # here, which is what a user hits seconds after running install.sh.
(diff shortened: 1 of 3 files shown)


gdevenyi:
Validated on real CI, not just locally. The `pr` lane cannot prove this PR — `rhel-8` is the only sub-floor entry in it, which is the whole reason #230 surfaced on a nightly — so I dispatched the affected nightly keys against this branch on my fork:

https://github.com/gdevenyi/uvr/actions/runs/30822267487 — **completed/success**

```
rocky-8        >>> expected failure confirmed: glibc 2.28 < 2.34
almalinux-8    >>> expected failure confirmed: glibc 2.28 < 2.34
oracle-8       >>> expected failure confirmed: glibc 2.28 < 2.34
opensuse-155   >>> expected failure confirmed: glibc 2.31 < 2.34
rocky-9        binary-smoke-ok  sysreqs-smoke-ok
Report matrix drift: skipped
```

I checked the markers rather than the exit codes, because exit 0 on its own would also be what a silently-skipped stage looks like — and "skipped instead of asserted" is precisely the failure mode this mechanism exists to avoid. Each of the four printed the confirmation, so the assertion ran and passed.

`rocky-9` is the control that matters: glibc exactly `2.34`, so it takes the normal path and installs R. The floor sits exactly on the RHEL 8 → 9 line, which makes `>= floor` load-bearing rather than decorative — an off-by-one there would have turned every 9-series distro into a false "expected failure" and looked green.

`Report matrix drift: skipped` is the same job that opened #230, staying quiet.

Combined with the earlier local runs against `ubuntu-2004` and `debian-11` (the two sub-floor entries not in this dispatch, both confirmed at glibc 2.31), all seven sub-floor entries are now accounted for.

gdevenyi:
One red check, and it is a CDN flake rather than this PR — but I cannot clear it myself (no rerun rights), so here is the evidence for whoever does.

`Test (musllinux-x86_64)` failed downloading R from Posit's CDN:

```
  Network error: error sending request for url (https://cdn.posit.co/r/musllinux_1_2/R-4.5.1-musllinux_1_2.tar.gz)
  client error (Connect)
  Connection reset by peer (os error 104)
```

Three reasons to read that as unrelated:

1. **It is a `ci.yml` job this PR does not touch.** The diff is `ci/distro-suite.sh`, `distro_matrix.json` and the design doc; nothing in it reaches the musl smoke test.
2. **#233 contains this exact commit plus its own, and that job passed there** — `Test (musllinux-x86_64): SUCCESS`. Same code, same runner image, different outcome.
3. The failure is `Connection reset by peer` on a CDN GET, i.e. transport, not anything uvr decided.

A rerun should clear it.

Separately, that traceback is a useful data point for the error-reporting question I raised on #218: this path *does* surface the underlying cause (`client error (Connect)`, `os error 104`), while `fetch_available_versions` renders only the outer `Network error: error sending request for url (…)` with nothing beneath it. So the gap I described there is per-call-site rather than universal — I will correct that comment.

nbafrank:
Approved. I checked this against the possibility that it was silencing real failures rather than fixing bookkeeping, and it's the latter — the four newly-covered entries are genuine members of the same sub-floor family as the rhel-8 case we already documented (rocky-8/almalinux-8/oracle-8 are RHEL 8 rebuilds at glibc 2.28, opensuse-155 is 2.31, floor is 2.34).

I verified the mirrored constants rather than taking them on trust: `R_GLIBC_FLOOR=2.34` matches `MANYLINUX_GLIBC_MIN` at `downloader.rs:147`, and the grepped substring matches the real error string at `downloader.rs:285` verbatim. Diff touches only `ci/`, the matrix JSON, and the design doc — no product code.

The part I like: the host-derived check is strictly *more* assertive than the per-entry `expect` blocks it replaces. Those could only say "uvr must refuse here"; this also fails if uvr ever *stops* refusing on a sub-floor host. That's a direction the old mechanism couldn't express at all.

One non-blocking note: `R_GLIBC_FLOOR` is a hand-copied literal mirroring the Rust constant. I traced the drift case and it fails loudly rather than silently (a host Rust newly refuses would fall into the normal install path and fail the ordinary success assertion), so the failure mode is right — but a cross-reference comment, or a test asserting the two are equal, would remove the manual sync entirely.

The one red job (`Test (musllinux-x86_64)`) is a live-CDN flake — "Connection reset by peer" mid-download, unrelated to this diff, and the same job passed clean on #233's later run over the same commits. Worth noting separately that integration tests hitting the real CDN will keep producing this class of false negative.

nbafrank:
Merged. Deriving the expectation from the host rather than from hand-authored per-entry notes is the right shape — #211 asserted it for three of seven sub-floor entries because only three had notes that happened to match, which is exactly the kind of gap that's invisible until a nightly goes red.

Issue #231, opened 2026-08-03
Alpine 3.20: R installs, then every source package fails — musl R's Makeconf needs gcc >= 14


gdevenyi:
Found by the nightly distro suite (#230). Not uvr's code, but uvr's users, and uvr currently gives them no way to understand it.

## What a user sees

`uvr r install 4.5.1` succeeds. R runs. Then any package with compiled code fails:

```
 x ERROR  Failed to install packages after add
   Failed to install rlang
   R CMD INSTALL failed for 'rlang' (exit 1):
   gcc: error: unrecognized command-line option '-std=gnu23'; did you mean '-std=gnu2x'?
   using C compiler: 'gcc (Alpine 13.2.1_git20240309) 13.2.1 20240309'
```

Nothing in that names the actual problem, and the install step that *caused* it succeeded several minutes earlier.

## Cause

R's `etc/Makeconf` records the compiler invocation its build host used. The two portable builds Posit publishes disagree — read straight off the CDN:

```
$ tar xzf R-4.5.1-musllinux_1_2.tar.gz  -O '*/lib/R/etc/Makeconf' | grep -E '^(CC|CXX) *='
CC = gcc -std=gnu23
CXX = g++ -std=gnu++17

$ tar xzf R-4.5.1-manylinux_2_34.tar.gz -O '*/lib/R/etc/Makeconf' | grep -E '^(CC|CXX) *='
CC = gcc
CXX = g++ -std=gnu++17
```

`-std=gnu23` landed in gcc 14. So the musl build requires gcc >= 14 on the *user's* machine, and the glibc build requires nothing. Matrix images:

| image | gcc | source builds |
|---|---|---|
| alpine:3.20 | 13.2.1 | **broken** |
| alpine:3.21 | 14.2.0 | fine |
| alpine:3.22 | 14.x | fine |

Any musl host with gcc < 14 is affected; Alpine 3.20 is just the one in the matrix.

## Where it belongs

Upstream, in [rstudio/r-builds](https://github.com/rstudio/r-builds) — the musl build environment recorded a compiler flag its consumers may not have, and the glibc one did not. Worth reporting there; a portable build should not pin a C standard newer than its oldest supported target.

## What uvr could do meanwhile

Nothing repairs the toolchain, but the silence is fixable and is most of the pain:

[...]

PR #233, opened 2026-08-03
ci: record Alpine 3.20's inability to compile R packages

gdevenyi:
**Stack position: merge second, after #232.** Built on that branch — until it lands, GitHub shows its commit in this diff. Closes the remaining failure in #230.

Unlike the other four failures in #230, this one is not CI bookkeeping. It is a real limitation users hit, and uvr currently gives them nothing to go on.

## What a user sees

`uvr r install 4.5.1` succeeds. R runs. Then anything with compiled code fails:

```
  Failed to install rlang
  R CMD INSTALL failed for 'rlang' (exit 1):
  gcc: error: unrecognized command-line option '-std=gnu23'; did you mean '-std=gnu2x'?
  using C compiler: 'gcc (Alpine 13.2.1_git20240309) 13.2.1 20240309'
```

The step that caused it succeeded several minutes earlier, and nothing in the message names the actual problem.

## Cause

R's `etc/Makeconf` records the compiler invocation its build host used. The two portable builds disagree — read straight off the CDN rather than inferred:

```
musllinux_1_2   CC = gcc -std=gnu23
manylinux_2_34  CC = gcc
```

`-std=gnu23` arrived in gcc 14. So the **musl** build requires gcc >= 14 on the user's machine and the glibc build requires nothing at all. That asymmetry is why only one lane went red:

| image | gcc | source builds |
|---|---|---|
| alpine:3.20 | 13.2.1 | **broken** |
| alpine:3.21 | 14.2.0 | fine |
| alpine:3.22 | 14.x | fine |

Any musl host with gcc < 14 is affected. Root cause is upstream in rstudio/r-builds — a portable build should not pin a C standard newer than its oldest supported target — so this PR records it and **#231** tracks it, including what uvr could do to stop being silent about it (`uvr doctor` can probe the installed R's `Makeconf` flags against the host compiler directly).

## The change

`expect` now understands the `sysreqs` stage as well as `r`, and alpine-320 carries:

```json
"expect": {
  "stage": "sysreqs",
  "message": "unrecognized command-line option '-std=gnu23'"
}
```

**Deliberately per-entry, not derived** — the opposite call from #232, and for a reason. The glibc floor is a host property that should self-maintain silently. This one *should* fail loudly the moment it stops being true, so that upstream fixing the Makeconf tells someone #231 can be closed, instead of the matrix quietly self-correcting and the issue rotting open.

## The coverage is not traded away

This entry exists for #30/#207 — Alpine sysreqs resolution — and that assertion still runs. The stage still reaches the common "did uvr skip the sysreqs check" check, and uvr still does the right thing right up to the compiler:

```
xml2 needs: libxml2-dev
> Running: apk add libxml2-dev
v System dependencies installed.
```

Only the compile step fails. That is the distinction worth keeping visible: uvr's sysreqs half works on Alpine 3.20; the toolchain is what does not.

## Verified

```
alpine-320  exit=0  expected failure confirmed: unrecognized command-line option '-std=gnu23'
alpine-321  exit=0  sysreqs-smoke-ok           <- unaffected, no expect block
ubuntu-2204 exit=0  binary-smoke-ok  sysreqs-smoke-ok
```

shellcheck clean, `sh -n` and `bash -n` clean.

https://claude.ai/code/session_01B2dAmERdjdAJr7xotiKQPB

### DIFF
--- a/ci/distro-suite.sh
+++ b/ci/distro-suite.sh
@@ -356,7 +356,26 @@ if ! skipped sysreqs; then
     "$UVR" init --here sysreqs-smoke --r-version "$R_LATEST"
     printf 'library(%s); cat("sysreqs-smoke-ok\\n")\n' "$SYSREQS_PKG" > check.R
 
-    if sysreqs_autoinstall_supported; then
+    if expect_fail sysreqs; then
+        # This host has a working toolchain that still cannot build R packages,
+        # because R's own Makeconf asks for a C standard its compiler predates.
+        # Asserting it keeps the limitation visible and checked; the sysreqs
+        # half is still verified by the common check below, which is the point
+        # — uvr resolves and installs the system dependency correctly here, and
+        # only the compile step fails.
+        if UVR_INSTALL_SYSREQS=1 "$UVR" add "$SYSREQS_PKG" --no-binary > add.log 2>&1; then
+            cat add.log
+            fail "the source build succeeded, but the matrix records this distro as
+       unable to compile R packages. If the toolchain moved or R's recorded
+       C standard changed, drop the 'expect' block from this entry."
+        fi
+        cat add.log
+        grep -qF "$EXPECT_FAIL_MESSAGE" add.log || fail \
+            "the source build failed, but not for the documented reason.
+       wanted: $EXPECT_FAIL_MESSAGE
+       Something else is broken here, or the known breakage has changed shape."
+        note "expected failure confirmed: $EXPECT_FAIL_MESSAGE"
+    elif sysreqs_autoinstall_supported; then
         UVR_INSTALL_SYSREQS=1 "$UVR" add "$SYSREQS_PKG" --no-binary 2>&1 | tee add.log
         "$UVR" run check.R
     else
--- a/ci/distro-suite.sh
+++ b/ci/distro-suite.sh
@@ -225,6 +225,38 @@ UVR="$DIST/uvr-$TARGET/uvr"
 note "install.sh would fetch: uvr-$TARGET.tar.gz"
 [ -x "$UVR" ] || fail "no release build for $TARGET under $DIST"
 
+# Whether the portable R builds can run here at all.
+#
+# Posit publishes them as manylinux_2_34, so a glibc host below 2.34 gets no R
+# — `MANYLINUX_GLIBC_MIN` in r_version/downloader.rs, and a different number
+# from both the 2.28 floor the uvr *binary* is built against (#212) and the
+# 2.28 `PPM_MANYLINUX_GLIBC_MIN` that gates binary *packages*. Three floors,
+# three meanings; only this one decides whether `uvr r install` can work.
+#
+# Derived from the host, deliberately. This used to be an `expect` block on
+# each entry, which meant it was only as complete as whoever remembered to add
+# one — and it was not: rocky-8, almalinux-8, oracle-8 and opensuse-155 are all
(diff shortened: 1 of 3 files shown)


gdevenyi:
Validated on real CI. `alpine-320` is nightly-lane, so the `pr` lane cannot exercise it either — dispatched against this branch on my fork:

https://github.com/gdevenyi/uvr/actions/runs/30822267487 — **completed/success**

```
alpine-320  >>> expected failure confirmed: unrecognized command-line option '-std=gnu23'
            >>> distro suite complete
```

Checked the marker rather than the exit code — exit 0 alone would look the same as a silently skipped stage, and the assertion firing is the point.

Two things this run confirms beyond "it is green":

**The message survives the trip.** It contains single quotes (`'-std=gnu23'`), and it reaches the container intact because it travels through `env:` rather than being interpolated into the shell line — the quoting hardening added in #211. Worth noting since a naive `-e "MSG=${{ ... }}"` would have broken here.

**The sysreqs coverage is intact.** This entry exists for #30/#207, and the stage still reaches the common "did uvr skip the sysreqs check" assertion. uvr does the right thing right up to the compiler:

```
xml2 needs: libxml2-dev
> Running: apk add libxml2-dev
v System dependencies installed.
```

So what is recorded here is narrow and accurate: uvr's sysreqs resolution works on Alpine 3.20, and the toolchain is what does not. `alpine-321` and `alpine-322` remain unmarked and pass normally.

nbafrank:
The technical claim checks out — I pulled both R 4.5.1 tarballs from the CDN and diffed `Makeconf` directly rather than trusting the excerpt:

```
musllinux_1_2:  CC = gcc -std=gnu23
manylinux_2_34: CC = gcc
```

So the musl build really does hardcode a flag gcc only learned in 14 (earlier gcc knows it as `-std=gnu2x`, which is exactly what the "did you mean" hint in the captured error says). Real upstream asymmetry, correctly diagnosed. The `expect_fail sysreqs` branch reuses #211's existing extension point cleanly and keeps the underlying sysreqs-resolution assertion running underneath, so we don't trade away the #30/#207 Alpine coverage to buy the expected-failure.

**One thing I want to name before this lands**, and it's about incentives rather than code: this PR is CI-only. Neither remediation from #231 — the `uvr doctor` compiler probe or rewriting the source-build error — is in the diff, which your PR body says plainly. The consequence is that merging it makes the nightly matrix fully green on alpine-320 while a real Alpine 3.20 user gets exactly the same bare `gcc: error: unrecognized command-line option '-std=gnu23'` as before. The matrix going green is the one piece of automated pressure that would otherwise keep pushing us to close that gap — your own design note says this "should fail loudly the moment it stops being true," and I want to make sure it doesn't also stop being loud about the part that's still broken for users.

I'm fine merging as CI hygiene. But I'd like #231 to stay explicitly open as user-facing work, and ideally a sense of whether you want to take the `doctor` probe or whether I should. "Matrix is green" must not come to mean "Alpine 3.20 users are covered."

Also: the design doc's stage-5 sysreqs narrative wasn't updated to mention the new branch (only the bugs table at the bottom) — #232 updated its stage section thoroughly, so this is just consistency.

nbafrank:
Merged. Thanks for chasing the root cause upstream rather than just marking it flaky — the Makeconf diff between the musl and glibc builds is the kind of finding that saves the next person a day.

#231 stays open as agreed. Happy to take the `uvr doctor` compiler probe myself if you'd rather stay on the CI side; say which you prefer.

Issue #238, opened 2026-08-06
`uvr` fails to find `gfortran`


AliSajid:
I am trying `uvr` out for a new bioconductor-based project. I am running into the following problem where a package that is compiling throws an error about not finding the `gfortran`.

Here are the logs:

```
󰄛 ❯ uvr add S4Vectors --bioc
~ S4Vectors [bioc] (updated)
› Installing 64 package(s): 58 binary · 6 from source
 ✗ ERROR  Failed to install packages after add
  Failed to install fields
  R CMD INSTALL failed for 'fields' (exit 1):
  /opt/homebrew/opt/llvm/bin/clang -I"/opt/homebrew/Cellar/r/4.6.1/lib/R/include" -DNDEBUG   -I/opt/homebrew/opt/llvm/include -I/opt/homebrew/include -I/opt/homebrew/opt/gettext/include -I/opt/homebrew/opt/readline/include -I/opt/homebrew/opt/xz/include    -fPIC  -g -O2 -fPIC -Wall -Wno-unknown-warning-option -c ExponentialUpperC.c -o ExponentialUpperC.o
  /opt/homebrew/opt/llvm/bin/clang -I"/opt/homebrew/Cellar/r/4.6.1/lib/R/include" -DNDEBUG   -I/opt/homebrew/opt/llvm/include -I/opt/homebrew/include -I/opt/homebrew/opt/gettext/include -I/opt/homebrew/opt/readline/include -I/opt/homebrew/opt/xz/include    -fPIC  -g -O2 -fPIC -Wall -Wno-unknown-warning-option -c RdistEarth.c -o RdistEarth.o
  /opt/homebrew/opt/llvm/bin/clang -I"/opt/homebrew/Cellar/r/4.6.1/lib/R/include" -DNDEBUG   -I/opt/homebrew/opt/llvm/include -I/opt/homebrew/include -I/opt/homebrew/opt/gettext/include -I/opt/homebrew/opt/readline/include -I/opt/homebrew/opt/xz/include    -fPIC  -g -O2 -fPIC -Wall -Wno-unknown-warning-option -c addToDiagC.c -o addToDiagC.o
[...]

PR #240, opened 2026-08-06
fix: diagnose a CRAN binary failing to load under a non-CRAN R

gdevenyi:
Refs #238.

## What actually failed there

Reading the full log in #238, `gfortran` is present and working — it compiled `fields`' own Fortran a few lines above the failure. What died is the **lazy-load of `dotCall64`**, a dependency that was installed as a **CRAN mac binary**:

```
dlopen(.../.uvr/library/dotCall64/libs/dotCall64.so, 0x0006): Library not loaded:
/Library/Frameworks/R.framework/Versions/4.6/Resources/lib/libgfortran.5.dylib
```

CRAN binaries hard-link CRAN's R.framework path for the Fortran runtime. The reporter runs **Homebrew R** (`/opt/homebrew/Cellar/r/4.6.1`), so that path doesn't exist. dlopen's "tried" list shows the loader's fallback — the running R's own `lib/` dir — which is why only Fortran-linking binaries break under Homebrew R while pure C/C++ binaries load fine.

The released binary routed this to the "missing Fortran toolchain / install CRAN gfortran" hint (wrong door: they have gfortran); current main's tightened matcher (`emutls_w` / `/opt/gfortran`) gives it **no hint at all**.

## Fix

A more specific `hint_for` branch, checked before the toolchain patterns: `Library not loaded` + `/Library/Frameworks/R.framework` → explain the CRAN-binary/non-CRAN-R mismatch, with three ways out, best first:

1. **Switch the project to a uvr-managed R** — `uvr r install <ver> && uvr r pin <ver>`. Verified against the real `R-4.6.1-macos-arm64.tar.gz` portable archive: it ships `lib/libgfortran.5.dylib` and `lib/libquadmath.0.dylib`, and R puts `$R_HOME/lib` on `DYLD_FALLBACK_LIBRARY_PATH` — the exact fallback path the reporter's error prints. CRAN binaries load under it, and it's the one remedy that is uvr's own mechanism.
2. Use CRAN R from cran.r-project.org/bin/macosx/.
3. Keep Homebrew R and `brew install gcc`, then symlink the libraries named in the error from `$(brew --prefix gcc)/lib/gcc/current/` into `$(R RHOME)/lib/`.

Two tests: the #238 message routes to the new hint (asserting the managed-R option leads, and not the toolchain hint), and a genuine missing-toolchain failure (`/opt/gfortran/...: No such file or directory`) keeps its existing hint.

## Not attempted here

The *mechanism* fix — e.g. having `uvr run`/install put Homebrew gcc's lib dir on `DYLD_FALLBACK_LIBRARY_PATH`, the mac analog of the `LD_LIBRARY_PATH` handling REnv already does on Linux — would make these binaries just work under Homebrew R too, but it changes loader behavior for every mac user and needs verification on actual macOS, which I can't do from this machine. Happy to file it as a follow-up if you think it's the right direction.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,15 @@ release page on GitHub. Issue numbers reference https://github.com/nbafrank/uvr/
 
 Pure tracking section — fixes and small features land here between tags.
 
+- **A CRAN mac binary failing to load under a non-CRAN R is now diagnosed
+  as what it is** (#238). When a package's compiled code cannot load because
+  it expects CRAN's R at `/Library/Frameworks/R.framework` (dlopen's
+  "Library not loaded"), the hint now explains the CRAN-binary/Homebrew-R
+  mismatch and both ways out — use CRAN R, or make the Fortran runtime
+  findable to the loader. Previously this failure either got no hint at
+  all or the "missing Fortran toolchain" one, which told users who already
+  had a working gfortran to go install one.
+
 ## v0.4.5 (2026-08-03)
 
 The community release. Four contributors filed thirteen pull requests in
--- a/crates/uvr/src/main.rs
+++ b/crates/uvr/src/main.rs
@@ -290,11 +290,13 @@ fn hint_for(msg: &str) -> Option<&'static str> {
         Some(
             "A package installed as a CRAN binary expects CRAN's R at \
              /Library/Frameworks/R.framework, and this R (Homebrew?) is not it, so the \
-             binary's compiled code cannot load. Either use CRAN R from \
-             https://cran.r-project.org/bin/macosx/, or keep this R and make the runtime \
-             findable: `brew install gcc`, then symlink the libraries named in the error \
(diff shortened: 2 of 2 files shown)


nbafrank:
Approved.

First, the red check: I pulled the failing `Distro (rocky 10 (rhel10))` job and it never reached your code — it died at "Getting action download info" with `Service Unavailable` / `Bad Gateway`, inside GitHub Actions' own action-resolution step. Same signature across #241's seven failures in the same window. Infra outage, not the PR.

The fix itself is right and the general match is the better call — keying on `library not loaded` + `/library/frameworks/r.framework` rather than hardcoding `libgfortran.5.dylib` means sibling cases like `libquadmath` are covered without another patch. Placing it before the `emutls_w`/`/opt/gfortran` branch matters, and your two tests pin both directions: the new hint leads with the uvr-native fix, and a genuinely missing toolchain still gets the toolchain hint rather than being swallowed.

Issue #170, opened 2026-07-14
cleanup: version normalization consistency (X.Y vs X.Y.0) across 4 sites


gdevenyi:
**Severity:** LOW (consistency)
**Component:** cross-cutting
**Labels:** `cleanup`, `good first issue`

## Problem

`X.Y` vs `X.Y.0` mismatch bites in 4 independent sites. Each one was fixed locally without coordinating on a shared normalizer, so they continue to drift.

## Site table

| Site | Problem | Symptom |
|---|---|---|
| `crates/uvr/src/commands/r_cmd/use_version.rs:20-23` | Writes `4.5` verbatim to `.r-version` (already tracked by #136) | `find_exact_version` finds no match |
| `crates/uvr-core/src/r_version/manager.rs:30` | `binary_for_version` does exact string match | Same `4.5` vs `4.5.0` mismatch |
| `crates/uvr/src/commands/r_cmd/list.rs:25` | `installed.contains(ver.as_str())` compares CDN version strings to dir names | Reports not-installed despite install existing |
| `crates/uvr-core/src/r_version/detector.rs:185-188` | `version_cmp` compares unequal-length Vecs (already tracked by #157) | `[4,5]` < `[4,5,0]` despite semver-equal |

## Suggested fix

Centralize on one normalizer (e.g. `normalize_ua_r_version` in `crates/uvr-core/src/r_version/downloader.rs`) used everywhere a version string is compared, written, or displayed. Apply at all four sites.

Could be done as a single PR closing #136, #157, and this issue together.

## Cross-references

- Related: #136, #157.
- Audit context: comprehensive codebase review.

nbafrank:
Fixed in [v0.4.1](https://github.com/nbafrank/uvr/releases/tag/v0.4.1).

PR #241, opened 2026-08-06
feat: add --install-dir to uvr r install

### LINKED ISSUE #89: allow for specifying R install directory in `uvr r install`
Users may want to choose their own directory for installing R versions with uvr.

We can do this If we add an arg such as `uvr r install --dir /path/to/install`, defaulting to the current install location

gdevenyi:
Closes #89.

Implements the flag as agreed in the thread: **scope (a)** (bsirak's pick — the flag sets the r-versions directory for this one `uvr r install` invocation) and **CLI wins over `UVR_R_INSTALL_DIR`** (uv convention, everyone agreed).

```sh
uvr r install 4.6.0 --install-dir /opt/shared/r-versions
```

## How

`RInstallArgs` gains `--install-dir <DIR>`; the override threads through `RManager::install` into `download_and_install_r`, which now picks the base as `flag → UVR_R_INSTALL_DIR → ~/.uvr/r-versions`. Everything downstream is untouched and applies to a custom root identically: partial-version resolution (#170), the broken-install detect-and-replace path, atomic staging rename (staging lives next to the target, so no cross-device surprises), and the post-install `R --version` probe. The existing `create_dir_all(staging_root)` already materialises a fresh custom root.

Discovery is deliberately unchanged: `r list` / pins / doctor find installs via `UVR_R_INSTALL_DIR`, and the flag's help text says exactly that, so nobody installs somewhere the rest of the tool isn't looking without being told the second step. (That broader flag/env pattern — `--cache-dir` etc. — stays with bsirak's follow-up RFC per the thread.)

Test: CLI help surface asserts the flag and its env-var relationship are advertised. An end-to-end install test would need a ~200 MB download, so verification of the plumbing rests on the type-checked single seam plus the unchanged downstream path.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ release page on GitHub. Issue numbers reference https://github.com/nbafrank/uvr/
 
 Pure tracking section — fixes and small features land here between tags.
 
+- **`uvr r install --install-dir <DIR>`** (#89) installs R into a chosen
+  r-versions directory for that one invocation, taking precedence over
+  `UVR_R_INSTALL_DIR` the way uv's flags beat their env vars. The env var
+  remains the way to make every command look there.
+
 ## v0.4.5 (2026-08-03)
 
 The community release. Four contributors filed thirteen pull requests in
--- a/crates/uvr-core/src/r_version/downloader.rs
+++ b/crates/uvr-core/src/r_version/downloader.rs
@@ -617,20 +617,28 @@ pub fn user_agent(info: &HostInfo) -> String {
     )
 }
 
-/// Download and extract R to `~/.uvr/r-versions/<version>/`.
+/// Download and extract R to `<r-versions>/<version>/`.
 ///
 /// `version` may be partial (`4.5`): it is resolved to the newest published
 /// matching version before install, so the install directory is always a
 /// full `X.Y.Z` that pin resolution and `r list --all` can match (#170).
+///
+/// The r-versions directory is `install_root` when given (`--install-dir`,
+/// which wins over the env var per uv convention, #89), else
+/// `UVR_R_INSTALL_DIR` / `~/.uvr/r-versions`.
 pub async fn download_and_install_r(
     client: &reqwest::Client,
     version: &str,
     platform: Platform,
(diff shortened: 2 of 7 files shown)


gdevenyi:
Copilot flagged that `--install-dir ''` reaches the downloader and fails with a message that does not name the flag. It cannot: clap's `PathBuf` parser refuses empty values at the argument layer ("a value is required"). I verified this against the built binary, and 441f0a9 adds a CLI test that pins the guarantee, so a parser change cannot quietly reopen the hole. No runtime guard is necessary.

gdevenyi:
Note on the red checks: the failures are the ongoing GitHub Actions outage (the jobs die in "Set up job" while they resolve `actions/checkout` — no uvr code runs). The full test suite is green locally. I will retrigger CI when GitHub reports Actions operational again.

nbafrank:
Approved.

The seven failures are the same GitHub Actions outage as #240 — every one dies at "Getting action download info" with `Service Unavailable`, and the `Select distros` / `Build static uvr` cancellations are downstream of it. Nothing here is your code.

The thing I specifically checked, given the history: `--distribution` on `r install` is deprecated-and-ignored, and `set_posit_distro_override` is *still* dead code on main today — defined, zero call sites, so a user whose distro is misdetected has no escape hatch despite the comment in `p3m.rs` promising one. **This PR doesn't repeat that.** `--install-dir` is wired end-to-end through a single seam: `RInstallArgs` → `install::run` → `RManager::install` → `download_and_install_r` → `install_dir.join(version)`, with the flag correctly winning over `UVR_R_INSTALL_DIR` per the convention agreed on #89. The broken-install replace, atomic staging, and post-install `R --version` probe all operate on the resolved dir, so none of them bypass a custom root.

The gap where `uninstall`/`list`/`use`/`pin` still resolve only via the env var is real but correctly scoped — you call it out in the flag's own help text, which is the right place for it.

One non-blocking suggestion: extracting the precedence into a pure `fn resolve_r_versions_base(flag, env) -> PathBuf` would make the rule directly testable without a download. Right now verification rests on the single seam being type-checked, which is true but untested.

Issue #205, opened 2026-07-30
Show per-package source + binary/source resolution before install (plan/dry-run)


pat-s:
### What

Before installing, uvr only prints an aggregate summary:

```
 INFO Using 1 binary-capable custom source(s); P3M suppressed.
> Installing 83 package(s): 80 binary - 3 from source
```

It's not visible **which** custom source was selected, nor **which** packages resolve to binary vs source and from **which** source — you have to inspect `uvr.lock` after the fact. Debugging an unexpected source build (e.g. a package pulled from GitHub via a transitive `Remotes:` instead of the binary repo) means reverse-engineering it from the lock.

### Request

Surface the resolution **before** the install runs (or via a `--dry-run`/plan flag):

- Name the binary-capable source(s) actually used, e.g. `rpkgs (https://cran.rpkgs.com/amd64/alpine323/latest)`, instead of `1 custom source(s)`.
- List each package with its resolved source + install kind, e.g.:
  ```
  openssl   2.4.2            binary  rpkgs
  RSQLite   3.53.3-4.9005    source  github (r-dbi/RSQLite)
  ```

This makes it obvious up front why something compiles and where each package comes from. A `-v`/`--verbose` gate is fine if you don't want it by default.

gdevenyi:
PR #242 implements the `-v` form of this request. Under `uvr sync -v`, at the point where the aggregate line prints and before any compile starts, you get:

- One row for each package: name, version, cached / binary / pure R / source, and the URL it resolved to.
- The name and base URL of each binary-capable custom source, under the "Using N binary-capable custom source(s)" line.

The rows use the runtime classification (the post-download tarball sniff), so they show what will happen, not the lock-time estimate. Default and `-q` output do not change.

A `--dry-run` / plan-only flag is not part of the PR. It is a larger design question, because an honest classification needs the download phase to run first.

nbafrank:
Shipped in #242 (@gdevenyi), merged today. `uvr sync -v` now names the binary-capable sources actually in use instead of "N custom source(s)", and prints one row per package with its resolved kind and the URL that served it — including the case where a binary plan fell back to the source tarball, which shows the URL that actually delivered the bytes rather than the one originally planned.

Gated behind `-v` as you said was acceptable, and wired through `EnvFilter::new` so `RUST_LOG` can't switch it on accidentally in CI.

PR #242, opened 2026-08-06
feat: show the per-package install plan under -v

gdevenyi:
Refs #205 — the verbose-gated slice of pat-s's ask, deliberately without new flags.

## What `-v` now shows

At the point the aggregate line prints (post-download tarball sniff, before anything compiles):

```
› Installing 83 package(s): 80 binary · 3 from source
  · openssl 2.4.2 — binary — https://cran.rpkgs.com/amd64/alpine323/latest/src/contrib/openssl_2.4.2.tgz
  · RSQLite 2.3.9 — source — https://api.github.com/repos/r-dbi/RSQLite/tarball/...
  ...
```

plus one row per cache-served package (`jsonlite 1.8.9 — cached`), and under `INFO Using 1 binary-capable custom source(s)` each source's name and base URL. Rows use the runtime classification (the same post-download sniff the aggregate line uses since the rpkgs.com work), so they say what will actually happen — an unexpected source build explains itself before the compiler starts, instead of being reverse-engineered from `uvr.lock`.

## Scope choices

- **Gate is the debug filter** — exactly what `-v` enables; default and `-q` output are byte-identical to before (asserted by the untouched existing output tests).
- **No `--dry-run` flag here**: pat-s's issue offers `-v` as an acceptable form, and a plan-without-installing mode is a bigger design call (it would need the download phase to run for honest classification). Left for the maintainer to decide separately.
- `cache_hit_count` became the `cache_hits` list it was counting — same value everywhere it was used (including the final "N% cache hit" line), plus the names `-v` needs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,15 @@ release page on GitHub. Issue numbers reference https://github.com/nbafrank/uvr/
 
 Pure tracking section — fixes and small features land here between tags.
 
+- **`uvr sync -v` shows the per-package install plan** (#205). Before
+  anything compiles, `-v` expands the aggregate "Installing 83: 80 binary ·
+  3 from source" into one row per package — name, version, cached/binary/
+  pure-R/source, and the URL it resolved to — and names each binary-capable
+  custom source behind the "Using N custom source(s)" line. An unexpected
+  source build now explains itself up front instead of being
+  reverse-engineered from `uvr.lock` afterwards. Default output is
+  unchanged.
+
 ## v0.4.5 (2026-08-03)
 
 The community release. Four contributors filed thirteen pull requests in
--- a/crates/uvr-core/src/registry/cran.rs
+++ b/crates/uvr-core/src/registry/cran.rs
@@ -317,6 +317,18 @@ impl CranRegistry {
     /// Returns true iff at least one entry in this registry has a `Built:`
     /// line matching the host. Used at sync time to decide whether this
     /// custom source contributes binaries.
+    /// Display identity: the source's name and its tarball base URL, for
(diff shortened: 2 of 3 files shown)


gdevenyi:
Copilot made a good catch: a binary plan whose download falls back to the source tarball printed the binary URL next to a `source` row. Fixed in db83c30 — the rows now zip the download results and show the URL that actually served the bytes (`fallback_url` when `is_binary && !used_binary`).

nbafrank:
Approved — this answers #205 in the form pat-s said was acceptable ("a `-v`/`--verbose` gate is fine if you don't want it by default"), and the two concerns I'd have raised are structurally handled rather than argued away.

On log spam and hot-path cost: `-v` maps through `EnvFilter::new`, not `from_default_env`, so `RUST_LOG` can't accidentally light these rows up in CI, and every new print sits behind `event_enabled!(DEBUG)` — the loops don't execute at all by default or under `-q`. That's the right shape.

The three-way `plans`/`detected_per_plan`/`results` zip made me look twice, but it rides an ordering invariant that already existed — the aggregate binary/source counts depend on the same one — so it's not new risk.

Nice catch on the follow-up commit: a binary plan whose download fell back to source would have printed the binary URL beside a `source` label. `if plan.is_binary && !result.used_binary { fallback_url }` is correct against `DownloadResult.used_binary`.

**One follow-up worth doing, not blocking the merge**: there are no tests on the new output. The URL-fallback bug shipped in the first commit and was caught by review rather than CI — a single `assert_cmd` test asserting the row shape (name/version/kind/URL, the cached row, the named custom source) against a fixture project would have caught it for free and protects the format from here on. Happy for that to land after.

PR #243, opened 2026-08-06
fix: honor UVR_LIBRARY everywhere the library is read

### LINKED ISSUE #97: Enable storing libraries outside project directories
Presumably this would be sensible to do via an env var, as is done by `UV_PROJECT_ENVIRONMENT` in `uv` and `RV_LIBRARY_DIR ` in `rv` - and `RENV_PATHS_LIBRARY` in `renv`.

The practical use case is that in some (mainly cloud) environments, the project folder is on a mapped/network drive, which makes packages symlinked to a global cache slow. Specifically, in Azure ML Studio the user code directory sits on mapped SMB-based storage, so we use these settings above (tested in uv and renv, not rv) to store the library on compute-local storage.

`rv` also provides an option to keep this path in rproject.toml (in which case authomatic scoping for R version/os/arch is lost - not sure if that is also the case for `RV_LIBRARY_DIR`). In the use cases I have seen the env var is the preferable option (it is machine specific, not fixed at project level) though there may be a case for both.

gdevenyi:
Closes #97.

## The gap

`UVR_LIBRARY` already exists — but only `uvr sync` honors it, and only as an **install target**. `uvr run`, `uvr activate`, doctor, and the `.Rprofile` snippet all hard-resolve `.uvr/library/`. So the exact workflow #97 describes (Azure ML mapped-SMB project dir, library redirected to compute-local disk) installs packages into a directory that no R session ever has on its search path.

## The fix — one rule, one seam each side

- **Rust side**: `Project::library_path()` resolves `UVR_LIBRARY` when set, else `.uvr/library/`. Every reader (`run`, `activate`, doctor, `ensure_library_dir`) already routes through it.
- **R side**: the `.Rprofile` snippet applies the same rule (`Sys.getenv("UVR_LIBRARY")`, else project-local), so RStudio/Positron sessions agree with the CLI. Existing projects pick this up automatically via the managed-block refresh on their next sync.

This matches the env-var precedent the issue cites (`UV_PROJECT_ENVIRONMENT`, `RENV_PATHS_LIBRARY`, `RV_LIBRARY_DIR`) and deliberately stays machine-local — no manifest/lockfile surface, per the reporter's own preference ("the env var is the preferable option; it is machine specific, not fixed at project level").

## What deliberately does not change

- `sync`'s precedence stays `--library` > `UVR_LIBRARY` > project-local.
- **Pruning never touches an override target** (they may be shared): `do_prune` keys off the override's *presence*, which the env var already sets — unchanged and still covered by its doc comment.
- `uvr run` outside a project keeps its existing fallback; the override is a project-scoped concept.

Tests: the consolidated env-var test asserts `library_path()` follows the override and falls back when unset/empty; a template test pins the `.Rprofile` contract so the two sides can't silently diverge.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,17 @@ release page on GitHub. Issue numbers reference https://github.com/nbafrank/uvr/
 
 Pure tracking section — fixes and small features land here between tags.
 
+- **`UVR_LIBRARY` now applies to the whole project, not just `uvr sync`**
+  (#97). Previously only sync honoured it as an install target, so setting
+  it sent packages to a directory that `uvr run`, `uvr activate`, doctor
+  and IDE sessions never looked at. All of them — including the `.Rprofile`
+  snippet, which refreshes on the next sync — now resolve the library
+  through the same rule: `UVR_LIBRARY` when set, else `.uvr/library/`.
+  This is the machine-local escape hatch for projects living on slow
+  network storage (Azure ML mapped drives and similar), matching
+  `UV_PROJECT_ENVIRONMENT` / `RENV_PATHS_LIBRARY` / `RV_LIBRARY_DIR`.
+  Pruning still never runs against an override target.
+
 ## v0.4.5 (2026-08-03)
 
 The community release. Four contributors filed thirteen pull requests in
--- a/crates/uvr-core/src/env_vars.rs
+++ b/crates/uvr-core/src/env_vars.rs
@@ -84,6 +84,8 @@ pub fn install_timeout() -> Option<String> {
 /// snippet), via `Project::library_path` (#97).
 /// Expects a valid absolute or relative directory path.
 /// Note: The CLI `--library` argument takes precedence over this variable.
+/// Defaults to the project-local `.uvr/library/` directory when unset or
+/// empty.
 pub fn library() -> Option<PathBuf> {
     read_env_var("UVR_LIBRARY").map(PathBuf::from)
 }
--- a/crates/uvr-core/src/env_vars.rs
+++ b/crates/uvr-core/src/env_vars.rs
@@ -78,10 +78,12 @@ pub fn install_timeout() -> Option<String> {
 
 /// UVR_LIBRARY
 ///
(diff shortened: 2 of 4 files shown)


gdevenyi:
Both Copilot points addressed in 6227257: the `.Rprofile` snippet now anchors the library with `normalizePath(lib, mustWork = FALSE)` at startup, so a relative `UVR_LIBRARY` survives a later `setwd()` instead of quietly re-anchoring; and the env var doc regains its explicit fallback note (project-local `.uvr/library/` when unset or empty).

gdevenyi:
Note on the red checks: the failures are the ongoing GitHub Actions outage (the jobs die in "Set up job" while they resolve `actions/checkout` — no uvr code runs). The full test suite is green locally. I will retrigger CI when GitHub reports Actions operational again.

nbafrank:
Approved.

The CI red is the same GitHub Actions outage that hit #240/#241 — all three jobs died at "Getting action download info / Service Unavailable" before any of your code ran. Rerun, nothing to fix.

The property I most wanted to check holds: with `UVR_LIBRARY` set and no `--library`, `library_override` is `Some(..)`, so `do_prune = prune && library_override.is_none()` is false and sync never prunes. That's the v0.4.3 CRITICAL finding staying fixed — pruning must never touch a library the user pointed elsewhere, because it may be shared. `read_env_var` filtering empty/whitespace values also means `UVR_LIBRARY=""` falls through to project-local rather than silently disabling pruning, which is the right call.

The `normalizePath(lib, mustWork = FALSE)` commit is a genuinely good catch — a relative `UVR_LIBRARY` going into `.libPaths()` unresolved would be silently dropped the moment an IDE calls `setwd()`, and pinning it absolute at startup before `dir.exists()` is exactly the right point.

One gap against the docstring's "every reader routes through it": `import.rs:216` still builds `.uvr/library` by hand and passes it to `ensure_companion_package`, so under `UVR_LIBRARY` an `import` installs the companion into an unreachable directory. Low severity — the next `sync` puts it in the right place, and import never prunes — but the claim is broader than the code. Fast follow-up or a docstring caveat, your call.

Issue #237, opened 2026-08-06
Linux container: warm sync 10x slower than cold at 100 packages — superlinear, cache path suspected


nbafrank:
## Symptom

In the Docker bench (first complete run, 30979470280), uvr's **warm** tier on tidyverse is ~10x slower than its own **cold** tier, and superlinear in package count:

| scenario | pkgs | uvr cold | uvr warm |
|---|---|---|---|
| ggplot2 | 17 | 4.81s | 2.36s |
| tidyverse | 100 | 6.08s | **65.39s** |

Warm-slower-than-cold is backwards (warm = caches populated, library cleared), and 17→100 packages scaling 2.4s→65s (6x pkgs, 28x time) points at something O(n²)-ish, not I/O.

For contrast, renv's warm tidyverse is 3.16s (symlink cache), and uvr's cold 6.08s shows the download+extract path is fine.

## What it is NOT

- Not the macOS/shared logic: the identical warm scenario on a laptop (same binary version, release build, `UVR_CACHE_DIR`/`UVR_PACKAGES_DIR` isolated) completes in **1.7s** with "100% cache hit".
- Not the Linux copy path per se: `clone_to_library` on Linux symlinks cache entries (package_cache.rs:269-289), which should be near-instant — *if* it's actually being hit.

## Leading hypothesis

The container's warm runs aren't actually hitting the installed-package cache — a cache-key/flavor mismatch or a `binary_cache_allowed` gate rerouting all 100 packages through the full per-tarball pipeline (sidecar sha256 read + `inspect_tarball` x2 + extract + verify + re-store) would produce exactly this shape on a slow runner. But the bench discarded the tools' own output, so the decisive "N% cache hit" line was never captured.

## Next step (instrumentation landed)

`BENCH_CMD_LOG` now captures each timed command's output into a `bench-cmds.log` artifact. The next benchmark run will show uvr's per-run cache-hit lines for the warm tier and settle the hypothesis.

## Why it matters

Warm sync is the everyday developer loop, and the honest benchmark table can't be published while uvr loses its own warm tier to a bug. (README refresh is holding on this.)

nbafrank:
Update from the instrumented run (31074711853): **the anomaly did not reproduce.**

With `bench-cmds.log` capturing uvr's own output, the warm tidyverse runs show:

```
v Installed 100 package(s) in 825ms
  100% cache hit
```

[...]

PR #245, opened 2026-08-06
fix: stop hashing the binary flavor into source cache keys

gdevenyi:
Refs #237 — found while trying to reproduce the warm-tier slowness. Full investigation notes going on the issue; the short version:

## The bug (code-provable, independent of #237's trigger)

- **Store** (`sync.rs`, the post-install cache write) passes the session's `binary_flavor` into `cache_key` for **every** package it caches — including PureR/Source-classified ones.
- **Lookup** (`package_cache::lookup_any`) probes source entries with `flavor = None`, per its own comment: *"a source build is compiled here and is not repo-specific."*

So on any Linux host with a flavour (every P3M-supported distro), a package that installs from source is stored under a key **no lookup can ever produce**. It is rebuilt from source on every warm sync, forever. macOS/Windows are untouched (`binary_repo_flavor()` is `None` there) — which is exactly why the laptop control in #237 showed 100% cache hits while the container didn't have to.

## Why this matters for #237's shape

Local reproduction in the bench image (same Dockerfile, R 4.5.3, run as root like CI) shows the healthy path is fine: ggplot2 warm 2.70s/2.69s, tidyverse warm **2.76s/2.78s at 100% cache hit**, all 100 entries `kind=binary`. So the CI 65s isn't intrinsic — but one window in which P3M serves source tarballs (the serving behavior ce75775 fixed for the comparison tools) during the warm tier's warmup would poison the cache with unreachable source-kind entries, and **every warm run after it re-pays the compiles** while the later cold tier, served binaries in a healthy window, looks fast. That reproduces the 10× warm-over-cold, the superlinearity (larger sets ⊃ more poisoned packages), and the fact it never self-heals. The `bench-cmds.log` from your next instrumented run will confirm or kill this: warm-tier plan lines reading "K from source" with a sub-100% hit rate is this mechanism; "100% cache hit" with a slow wall clock is something environmental instead.

## The fix

The invariant moves into `cache_key` itself: flavour is hashed only when `is_binary` — store and lookup agree by construction, and no call site can reintroduce the split. macOS/Windows and source keys are byte-identical to before; only Linux binary entries carry the flavour, exactly the ones #175 introduced it for. Existing mis-keyed source entries were unreachable anyway and are simply superseded (they'll age out via `uvr cache clean`).

Regression test pins `key(source, Some(flavor)) == key(source, None)` alongside the existing flavour-separation tests for binaries.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,16 @@ release page on GitHub. Issue numbers reference https://github.com/nbafrank/uvr/
 
 Pure tracking section — fixes and small features land here between tags.
 
+- **Source-built packages cached on Linux are now found again by later
+  syncs** (#237). The cache key hashed the binary-repo flavour into every
+  entry it stored, but lookups probe source entries without one — so on
+  any Linux host with a binary repo (a flavour), every package that
+  installed from source landed in the cache under a key no lookup could
+  produce, and was rebuilt from source on every warm sync, forever. The
+  flavour is now part of the key only for binary entries, which is the
+  rule lookups always assumed. Existing mis-keyed source entries were
+  unreachable anyway and are simply superseded.
+
 ## v0.4.5 (2026-08-03)
 
 The community release. Four contributors filed thirteen pull requests in
--- a/crates/uvr-core/src/installer/package_cache.rs
+++ b/crates/uvr-core/src/installer/package_cache.rs
@@ -83,10 +83,16 @@ pub fn cache_key(
     // artifacts linking different libraries, and only one of them loads on a
     // given host (#175) — so they must not share a cache entry.
     //
-    // Only hashed when present, which keeps macOS/Windows and source-built
-    // keys byte-identical to before; only Linux binary entries are
-    // re-keyed, and those are exactly the ones that were ambiguous.
-    if let Some(flavor) = binary_flavor {
+    // Only hashed for *binary* entries: a source build is compiled on this
+    // host and is not repo-specific. Enforced here, not at call sites —
+    // the store path passed the session's flavour unconditionally while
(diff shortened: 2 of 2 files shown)


nbafrank:
Approved — and this bug was worse than the title suggests, which is worth stating plainly: pre-fix, `store` hashed `binary_flavor` for *every* install including source builds, while `lookup_any` always probed the source key with `flavor = None`. So on flavored Linux **no source entry was ever reachable** — every warm sync recompiled from scratch, forever, writing cache entries nothing could ever read. Gating on `(is_binary, Some(flavor))` makes store and lookup agree by construction, and `a_source_entry_ignores_the_binary_flavor` pins it.

No functioning entry is invalidated: binary keys are byte-identical before and after, and the source entries this "changes" were already write-only. Worth adding a pointer to `uvr cache clean` in the CHANGELOG line, since the orphans under the old scheme are now permanent dead weight that nothing reclaims.

One thing I'd like acknowledged in a comment rather than changed in code: post-fix a Linux source key is `{checksum, r_minor, "source", ARCH-OS}`, and `std::env::consts::OS` is `"linux"` for everything — musl and glibc are indistinguishable there. On a single host that's fine, and your "not repo-specific" reasoning holds. But `UVR_PACKAGES_DIR` exists, and #243 is actively encouraging shared/network storage patterns; if someone ever points the package cache at storage shared by heterogeneous nodes, a source-compiled `.so` from a musl node could be served to a glibc one. Bounded failure (the loader errors at `library()`, same class as the binary-flavor mismatch the module docs already accept), but a one-sentence "shared cache directories across heterogeneous hosts are out of scope" would save the next person a debugging session.

PR #254, opened 2026-08-17
Update R version management and run scripts support

dpastoor:
I am not sure how this table is generated, however all these features marked as not available for rv have been available for months/years, so would appreciate this being properly reflected :-)

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,14 +78,14 @@ If you are happy with renv + rig, that is a perfectly good setup. `uvr` is for p
 |--------------------------------|-----|------|-----|-----|-----|------|
 | Declarative manifest           | Y   | Y†   | Y†  | Y   | -   | Y    |
 | Lockfile                       | Y   | Y    | Y   | Y   | -   | Y    |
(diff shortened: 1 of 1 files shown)


dpastoor:
in looking more, i am guessing some of this table may have come from when uvr was initially created, and by working off our docs (which have not always detailed some of the technical capabilities). Some features like rv run are newer and pretty quiet doc wise so not surprising it wasn't picked up. Congrats on the project - the more the merrier around making dep management in R feasible :-)

nbafrank:
Hi @dpastoor , I wanted to send you a note to thank you for 1) your work in rv and 2) your help with uvr. Thanks for taking the time to go through the README constructively and collaboratively :) I am relatively new to this and in this process I am making mistakes. Thank you for your help in fixing one and make comparisons more accurate and congrats for your work too! Happy to collab in the future!

p.s. I split the R version row in two since rv seems to pin among installed R versions rather than install R itself, but correct me if I misread and I'll fix it right away.

PR #257, opened 2026-08-24
[enhancement] opt-in IDE integration and CI/headless controls

### LINKED ISSUE #176: [enhancement] `--no-ide` flag
Thank you so much for this awesome tool! I'm using `uvr` now to manage dependencies in docker image builds, and noticed there's no workaround `ensure_positron_settings()`, which writes `.vscode/settings.json` for the positron IDE. Would it be possible to provide a `--no-ide` flag for `uvr init` and `uvr sync` to opt-out of this?

Happy to offer a PR if you're open to it ;)

EDIT: related to #206 (baked-in IDE assumption with other side effects)

### LINKED ISSUE #206: Headless/CI mode: skip companion package + IDE scaffolding + system-R hint
### Summary

Add a headless/CI mode that skips the interactive/IDE-oriented scaffolding. In CI (ephemeral container, system R, no editor) all of it is pure overhead and log noise.

Consolidates #204 (the `.r-version` hint is one symptom of the same scaffolding).

### What CI doesn't want

Currently `uvr init`/`sync` always:

1. **Installs the `uvr` companion R package** into the project library. It's not a project dependency (not in `uvr.toml`) — uvr injects it. In a headless run it's an extra source build and an extra package in the (e.g. `R CMD check`) library, with no benefit since nothing calls `uvr::add()`/`uvr::lock()` from R.
2. Writes the **`.Rprofile` block**, **`.gitignore` entry**, and **Positron/IDE config**.
3. Prints, on every `sync`/`add`:
   ```
   - No .r-version pin — IDE config bound to system R at /usr/local/bin/R. Run `uvr r install <ver>` then `uvr r pin <ver>` for project-pinned R.
   ```
   Not silenced by `-q`, by `uvr r pin` (writing `.r-version`), or by redirecting stderr (it's on stdout). Using the image's system R is intentional in CI.

### Request

One switch — e.g. `UVR_HEADLESS=1` (and/or `uvr init --bare` / `--no-companion`) — that disables all of the above: no companion package, no `.Rprofile`/`.gitignore`/IDE files, no system-R hint. Just resolve + install into the library.

Context: CI on Alpine using a system R, installing binaries from a CRAN-like repo; `uvr sync` is otherwise working great.

rmvegasm:
# Make IDE integration opt-in and add CI/headless controls

Fixes #176, Fixes #206.

## Problem

`uvr init` / `sync` / `import` assumed an RStudio/Positron workflow was always in
use, so they unconditionally wrote IDE scaffolding and installed the companion
R package. In CI/containers this is unwanted side effects and log noise.

## What changed

- IDE config is now **opt-in**:
  - `.vscode/settings.json` (and the "No `.r-version` pin" IDE hint) are only
    produced when an IDE is detected or forced.
  - Detection reads `POSITRON=1` (Positron) and `RSTUDIO=1` (RStudio).
  - `.Rprofile` is still written by default — it's the library wiring any R
    session started from the project root needs, not IDE config. `--unattended`
    and `--bare` drop it; the library stays reachable through `uvr run` /
    `R_LIBS_USER`.
- New flags:
  - `--ide=positron|rstudio` — force IDE behavior (scoped to `init` / `sync` /
    `import`).
  - `--no-ide` — disable IDE config/messages (scoped the same way).
  - `--no-companion` (env `UVR_NO_COMPANION=1`) — skip the companion R package.
  - `--unattended` (env `UVR_UNATTENDED=1`) — `--no-ide` + `--no-companion` +
    no working-tree writes (`.Rprofile`, `.gitignore`, shims, `.Rbuildignore`).
  - `uvr init --bare` — `uvr.toml`, `.uvr/library/`, and a protective
    `.gitignore`; no `.Rprofile`, shims, IDE config, or companion. Persists via
    `[project] bare = true`, and later `uvr sync` keeps the project bare.

## Behavior summary

| Mode | `.Rprofile` | `.gitignore` | `.vscode` | companion |
|---|---|---|---|---|
| default (no IDE detected) | yes | yes | no | yes |
| `POSITRON=1` / `--ide=positron` | yes | yes | yes | yes |
| `RSTUDIO=1` / `--ide=rstudio` | yes | yes | no | yes |
| `--no-ide` | yes | yes | no | yes |
| `--no-companion` | yes | yes | per IDE | no |
| `--unattended` | no | no | no | no |
| `init --bare` | no | yes | no | no |

Activation shims follow the `.Rprofile` column. `.Rbuildignore` follows it too,
for R package source trees (DESCRIPTION with a `Package:` field).

## Tests

Full workspace suite passes: 180 uvr unit tests, 72 CLI tests, 515 uvr-core
tests (network-only suites remain `#[ignore]`d).


gdevenyi (commented):
Reviewed against `main` (e589f03). I built the branch, ran the full suite, clippy and rustfmt.

**What checks out**

- 180 + 71 + 515 tests pass locally with R present, matching the PR description.
- `cargo clippy --workspace --all-targets -- -D warnings` is clean.
- The `.r-version` hint from #206 item 3 is correctly gated. It fires inside `resolve_project_r_binary`, whose only caller is `ensure_positron_settings` (`init.rs:258`), so gating that call on `ide.is_positron()` silences the hint transitively. Worth noting in the PR body, because it isn't visible in the diff.
- `--no-companion` reaches `uvr add` too: `add.rs:338` → `install_from_lockfile` → `install_from_lockfile_with_r`, which now reads `companion_wanted`. Threading it as an env var rather than a parameter is what makes that work.

**Blocking: `cargo fmt --all -- --check` fails**

CI runs it at `.github/workflows/ci.yml:62`, so this fails before the tests run. Nine hunks:

```
crates/uvr-core/src/env_vars.rs:114
crates/uvr/src/commands/import.rs:239
crates/uvr/src/commands/init.rs:113
crates/uvr/src/commands/sync.rs:166, 207, 3056
crates/uvr/src/ide.rs:132, 144
crates/uvr/src/main.rs:140
```

`cargo fmt --all` fixes all of them. The PR shows no check runs at all, so I suspect workflow approval for a first-time contributor is still pending and you have not seen this yet.

**Main design question: `--unattended` still writes into the working tree**

Built the branch and ran it:

```
$ git init -q . && uvr init --here proj --unattended
$ git status --porcelain
?? .Rprofile
?? .gitignore
?? uvr.toml
```

#206 lists the `.Rprofile` block and the `.gitignore` entry as item 2 of what CI does not want, alongside the companion package and the IDE config. Right now `--unattended` covers items 1 and 3 but not 2, so "Fixes #206" is partial.

The rationale in `ide.rs` — `.Rprofile` is library wiring, not IDE config — is sound for an interactive terminal. It is weaker in CI, where the point is that uvr must not modify a checked-out repository. A job that runs `git diff --exit-code`, or that builds a Docker layer, now carries two files it did not ask for. The library is already reachable through `uvr run` and `R_LIBS_USER`.

`--bare` does drop them, but it is a persistent property of the project written into `uvr.toml`, so it cannot be the answer for a repo that is interactive locally and headless in CI — which is the common case, and the one pat-s described. I would either fold `.Rprofile`/`.gitignore` into `--unattended` (making `--bare` just the persisted form of the same thing), or say explicitly in the docs that `--unattended` deliberately keeps them and close #206 as partially addressed. Your call on which, but the current state answers the issue only halfway.

**Smaller items**

1. `env_vars.rs` — the doc comment on `truthy` says "Accepts the same truthy values as `UVR_NO_BINARY`", but it accepts `on`/`ON` and `UVR_NO_BINARY` does not (`sync.rs:1769`, `1|true|yes|TRUE|YES`). Either the comment or the match arm is wrong.

   That is now the fourth hand-rolled copy of this match on this branch — `source_installs_forced` (`sync.rs:1769`), `cache_lookup_disabled` (`sync.rs:1780`) and `sysreqs_install_enabled` (`sync.rs:1794`) are the others, and all three agree with each other and disagree with the new one. Since this PR is already adding the shared helper, making it `pub` in `env_vars` and having the three `sync.rs` helpers call it would leave one definition instead of four, and the superset would be a deliberate widening rather than an accident.

2. `--bare` skips `.gitignore`, and `write_gitignore` (`init.rs:451`) is what ignores `/.uvr/library/`. So `uvr init --bare` inside a git repo leaves an installed R library fully untracked-and-visible, ready to be committed by `git add -A`. Was that intended? Ignoring the library costs nothing even in a bare project, and the failure mode is a multi-hundred-megabyte commit.

3. `--ide` is `global = true` but only `init`, `sync` and `import` consume it. `uvr add --ide=positron` parses and does nothing. Not harmful, but a scoped flag would not need explaining later.

4. Minor: `uvr activate` in a bare project regenerates `.uvr/activate*`, so `bare` is not sticky there. Correct behaviour — the user asked for it explicitly — but the `--bare` help text says "no activation shims" without qualification.

Nothing here is hard to fix and the shape of the change is right. The fmt failure and the `.Rprofile`/`--unattended` question are the two I would settle before merge.

rmvegasm:
Thanks for the detailed review, I amended the PR following your comments.
Removed the widening for `truthy` values, they are now collapsed into a single
helper but keeping the original set. Also updated the original PR desription.

## Updated behaviour

IDE integration is now opt-in, and CI/headless mode no longer touches the
working tree.

- IDE config (`.vscode/settings.json` and the IDE-oriented hints) is written
  only when an IDE is detected (`POSITRON=1` / `RSTUDIO=1`) or forced with
  `--ide=positron|rstudio`. With no IDE detected the default writes no
  `.vscode`. `--no-ide` suppresses it; both flags are now scoped to
  `init` / `sync` / `import`.
- `--unattended` / `UVR_UNATTENDED=1` disables the companion package, IDE
  config, and every working-tree write (`.Rprofile`, `.gitignore`, activation
  shims, `.Rbuildignore`). On `init`/`import` only `uvr.toml` and
  `.uvr/library/` are written; on `sync` only the library. The library is
  reached through `uvr run` or `R_LIBS_USER`.
- `uvr init --bare` is the persistent form for interactive use: `uvr.toml`,
  `.uvr/library/`, and a protective `.gitignore` (so the library can't be
  committed) — no `.Rprofile`, shims, IDE config, or companion.

## Coverage of #206

1. **Companion package** — disabled by `--unattended` / `--no-companion` /
   `--bare`. Covered.
2. **`.Rprofile` / `.gitignore` / IDE config** — `--unattended` writes none of
   them (and also skips the activation shims and `.Rbuildignore`). Covered.
3. **`.r-version` hint** — the hint lives in `resolve_project_r_binary`,
   whose only caller is `ensure_positron_settings`, so gating that call on
   `ide.is_positron()` silences it transitively. `--unattended`, `--no-ide`,
   `--bare`, and the default (no IDE) all resolve to `Ide::None` and never
   print it. Covered — note this is invisible in the diff.

`--unattended` therefore satisfies the issue's request: no companion, no
`.Rprofile`/`.gitignore`/IDE files, no system-R hint — just resolve and
install into the library. The switch is named `UVR_UNATTENDED` rather than
the `UVR_HEADLESS` spelling suggested in the issue.

## Coverage of #176

`--no-ide` now exists on `init`/`sync`/`import`. The default is stricter than
the flag alone: with no IDE detected, no `.vscode` is written at all, so
`--no-ide` is only needed to override a *detected* IDE.

## Notes

- `--bare` still writes `.gitignore` (protective), while `--unattended` writes
  nothing — bare is explicit interactive setup; unattended must leave a
  checked-out tree byte-identical.
- `--ide` is no longer global: `uvr add --ide=…` is rejected at parse time.
- Internal: the four hand-rolled truthy parsers were collapsed into one shared
  helper (`env_vars::truthy`), keeping the exact pre-existing accepted set.

crates/uvr/tests/cli_tests.rs
```diff
@@ -1519,3 +1519,138 @@ fn test_an_unsupported_r_pin_in_a_header_is_reported_not_swallowed() {
         .stdout(predicate::str::contains("RAN"))
         .stderr(predicate::str::contains("does not honour yet"));
 }
+
+// ─── IDE-mode scaffolding ─────────────────────────────────────────
+
+#[test]
+fn test_init_default_writes_rprofile_but_no_ide_config() {
```
gdevenyi, line 1526: `uvr_cmd()` passes the whole parent environment to the binary, so these tests change result with the environment of the person who runs them:

```
$ POSITRON=1 cargo test -p uvr --test cli_tests
test result: FAILED. 70 passed; 2 failed
  (this test, and test_init_rstudio_env_writes_rprofile_but_no_vscode,
   because POSITRON takes precedence over RSTUDIO)

$ UVR_UNATTENDED=1 cargo test -p uvr --test cli_tests
test result: FAILED. 63 passed; 9 failed
  (the 7 new test_init_* tests other than test_init_unattended_*,
   plus the older test_init_writes_activation_shims and
   test_activate_write_shim_restores_deleted_shims)
```

One change in `uvr_cmd()` (line 6, outside the diff) fixes all of them. The tests that set `.env("POSITRON", "1")` still work, because a later `.env()` overrides an earlier `.env_remove()`:

```rust
fn uvr_cmd() -> Command {
    let mut cmd = Command::cargo_bin("uvr").unwrap();
    for var in ["POSITRON", "RSTUDIO", "UVR_UNATTENDED", "UVR_NO_COMPANION"] {
        cmd.env_remove(var);
    }
    cmd
}
```

With this change, the full CLI suite passes under both `POSITRON=1` and `UVR_UNATTENDED=1` (72 passed).

crates/uvr/src/ide.rs
```diff
@@ -0,0 +1,159 @@
...
+            &[("POSITRON", Some("1")), ("UVR_UNATTENDED", Some("1"))],
+            || {
+                assert_eq!(Ide::resolve(None, false, false), Ide::None);
+            },
+        );
+
+        // No override: env detection wins.
+        with_env(&[("RSTUDIO", Some("1"))], || {
+            assert_eq!(Ide::resolve(None, false, false), Ide::Rstudio);
+        });
```
gdevenyi, line 157: This block has the same problem. From a Positron terminal `detect()` returns `Positron`, and with `UVR_UNATTENDED=1` set `resolve()` returns `None`:

```
$ POSITRON=1 cargo test -p uvr --bin uvr ide::
  left: Positron
 right: Rstudio

$ UVR_UNATTENDED=1 cargo test -p uvr --bin uvr ide::
  left: None
 right: Rstudio
```

Clear both variables in this block. This is the `rustfmt` form, and it passes under both variables:

```suggestion
        with_env(
            &[
                ("RSTUDIO", Some("1")),
                ("POSITRON", None),
                ("UVR_UNATTENDED", None),
            ],
            || {
                assert_eq!(Ide::resolve(None, false, false), Ide::Rstudio);
            },
        );
```

crates/uvr-core/src/manifest.rs
```diff
@@ -54,6 +54,13 @@ pub struct ProjectMeta {
 
     #[serde(default)]
     pub description: Option<String>,
+
+    /// Created with `uvr init --bare`: the project ships only `uvr.toml`
+    /// and `.uvr/library/`. `.Rprofile`, `.gitignore`, activation shims,
+    /// IDE config, and the companion package are all skipped, and the
+    /// library is reachable through `uvr run` only.
```
gdevenyi, line 61: This comment is from before the `.gitignore` change. A bare project now gets the protective `.gitignore`, so the text disagrees with `init.rs` and the `--bare` help text:

```suggestion
    /// Created with `uvr init --bare`: the project ships only `uvr.toml`,
    /// `.uvr/library/`, and a protective `.gitignore`. `.Rprofile`,
    /// activation shims, IDE config, and the companion package are all
    /// skipped, and the library is reachable through `uvr run`.
```

crates/uvr/src/cli.rs
```diff
@@ -116,6 +126,21 @@ pub struct InitArgs {
     /// R version constraint, e.g. ">=4.3.0"
     #[arg(long = "r-version", value_name = "CONSTRAINT")]
     pub r_version: Option<String>,
+
+    /// Assume a specific IDE for config generation (positron | rstudio)
+    #[arg(long, value_enum, value_name = "IDE", conflicts_with = "unattended")]
```
gdevenyi, line 131: `conflicts_with = "unattended"` works only when `--unattended` comes after the subcommand. When the global flag comes first, clap accepts the pair:

```
$ uvr init --here p2 --ide positron --unattended
error: the argument '--ide <IDE>' cannot be used with '--unattended'

$ uvr --unattended init --here p1 --ide positron
v Initialized project p1
```

`UVR_UNATTENDED=1 uvr init --ide positron` is also accepted. The result is correct in all three cases, because the write gates read `UVR_UNATTENDED`, so `--ide` has no effect. But the parse-time error appears for only one of the three spellings.

I would remove the three `conflicts_with = "unattended"` (here, and at lines 244 and 403) and let `--unattended` win, as it already does at runtime. The other option is to keep them and accept the gap.

crates/uvr/src/cli.rs
```diff
@@ -116,6 +126,21 @@ pub struct InitArgs {
...
+
+    /// Disable IDE-specific config files and messages
+    #[arg(long, conflicts_with = "ide")]
+    pub no_ide: bool,
+
+    /// Bare project: `uvr.toml` + `.uvr/library/` + a protective
+    /// `.gitignore`. No `.Rprofile`, activation shims (until you run
+    /// `uvr activate`), IDE config, or companion package — the library is
+    /// reachable through `uvr run` only.
+    #[arg(long)]
```
gdevenyi, line 142: `--bare` skips the Positron settings, so `uvr init --here p --bare --ide positron` accepts `--ide` and ignores it (it writes only `uvr.toml`, `.uvr/` and `.gitignore`). A conflict rejects the combination, as `--no-ide` already does:

```suggestion
    #[arg(long, conflicts_with = "ide")]
```

crates/uvr/src/ide.rs
```diff
@@ -0,0 +1,159 @@
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum Ide {
+    /// No IDE — don't write IDE config or print IDE hints.
+    None,
+    /// Positron — write `.vscode/settings.json` (`positron.r.*`, `r.rterm`,
+    /// `r.rpath`).
+    Positron,
+    /// RStudio — no `.vscode` config today; `.Rprofile` is already written
+    /// for every mode.
+    Rstudio,
```
gdevenyi, line 24: Question, not blocking: `Ide::Rstudio` has no behaviour of its own. `is_positron()` is the only thing that reads an `Ide`, so `Rstudio` and `None` act the same everywhere, and `--ide=rstudio` has the same effect as `--no-ide`.

Do you plan RStudio-specific config later? If not, a single `positron: bool` removes the `Rstudio` variant, `IdeArg`, and the `RSTUDIO` check in `detect()`. If yes, it would help to say in the help text that `--ide=rstudio` currently only overrides a detected Positron.

crates/uvr/src/ide.rs
```diff
@@ -0,0 +1,159 @@
...
+
+    /// Combine explicit CLI overrides with environment detection.
+    ///
+    /// Precedence: `--ide` > (`--no-ide` | `--unattended` | `UVR_UNATTENDED=1`)
+    /// > detected `POSITRON=1` / `RSTUDIO=1` > `None`.
+    pub fn resolve(cli: Option<Ide>, no_ide: bool, unattended: bool) -> Self {
+        if let Some(ide) = cli {
+            return ide;
+        }
+        if no_ide || unattended || uvr_core::env_vars::unattended() {
```
gdevenyi, line 48: Nit: `main.rs` sets `UVR_UNATTENDED=1` before it calls `resolve()`, so the `unattended` parameter always has the same value as `uvr_core::env_vars::unattended()` here. You can remove the parameter and keep the env check.

gdevenyi (commented):
Re-reviewed at 38913fe. Thanks for the changes. Both points from my last review are closed.

**What checks out**

- `cargo fmt --all -- --check` is clean, and `cargo clippy --all-targets --all-features -- -D warnings` is clean. The full suite passes: 180 + 72 + 515.
- `--unattended` now writes only `uvr.toml` and `.uvr/library/`. `uvr --unattended init --here p` leaves only those two in the directory, and `test_init_unattended_writes_only_manifest_and_library` covers it.
- `--bare` now writes the protective `.gitignore`, and the test checks for `/.uvr/library/`.
- `truthy` keeps the old accepted set. `read_env_var` only adds the empty-value filter, so `UVR_NO_BINARY`, `UVR_IGNORE_CACHE` and `UVR_INSTALL_SYSREQS` work as before.
- `--ide` is scoped now, and `uvr add --ide=positron` fails at parse time.
- The detection uses a real signal: Positron sets `POSITRON=1` in its terminal environment (`src/vs/platform/terminal/node/terminalEnvironment.ts` in posit-dev/positron).
- A test merge with current `main` builds and passes. Only `CHANGELOG.md` conflicts (the new #261 and #271 entries).

**Should fix: the new tests depend on the environment of the person who runs them**

Both the unit test and the CLI tests read `POSITRON`, `RSTUDIO` and `UVR_UNATTENDED` from the parent environment. From a Positron terminal, three tests fail. With `UVR_UNATTENDED=1` exported (for example by a CI job that sets it globally, which is what this PR asks CI users to do), ten fail, and two of them are older tests. Details and tested fixes are inline.

**Smaller items** are inline: a stale doc comment on `bare`, two gaps in the clap conflicts, and a question about `Ide::Rstudio`.

novica:
A case #257 doesn't fully cover: a repo that **commits** `.vscode/settings.json` and `.Rprofile`. It's a project
template, where the shared editor settings (formatter, format-on-save) are the point of tracking that file.

On uvr 0.4.6, running `uvr lock` then `uvr sync` from a plain terminal left the working tree dirty:

```diff
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -6,5 +6,14 @@
   "[r]": {
     "editor.defaultFormatter": "Posit.air-vscode",
     "editor.formatOnSave": true
-  }
+  },
+  "positron.r.customBinaries": [
+    "/home/<user>/.uvr/r-versions/4.6.1/bin/R"
+  ],
+  "positron.r.customRootFolders": [
+    "/home/<user>/.uvr/r-versions"
+  ],
+  "positron.r.interpreters.default": "/home/<user>/.uvr/r-versions/4.6.1/bin/R",
+  "r.rpath.linux": "/home/<user>/.uvr/r-versions/4.6.1/bin/R",
+  "r.rterm.linux": "/home/<user>/.uvr/r-versions/4.6.1/bin/R"
 }
```

plus a refreshed `# >>> uvr >>>` block in `.Rprofile`, because the snippet had changed in a newer uvr release.

Three things that remain after #257, as I read it:

1. **Absolute per-user paths in a shared file.** With `POSITRON=1` / `--ide=positron`, uvr still writes
   `/home/<user>/...` into `.vscode/settings.json`. In a tracked file, that's one developer's machine state committed
   for everyone, and it churns between contributors. Could uvr skip the write, or warn, when `.vscode/settings.json` is
   tracked by git (`git ls-files --error-unmatch`)? Another option is writing machine-specific keys to user-level
   settings instead of the workspace file.
2. **`.Rprofile` block refresh on uvr upgrades.** `ensure_rprofile` rewrites the block whenever `RPROFILE_SNIPPET`
   differs from what's on disk. Every uvr release that touches the snippet then dirties the tree for every
   contributor, one at a time, as each person upgrades. Rewriting only when the block is missing, or only on
   `uvr init`, would avoid that.
3. **`sync --frozen` writes before it checks.** In `sync.rs`, `run_inner` calls `ensure_rprofile` and
   `ensure_positron_settings` *before* the `if frozen` staleness check. A frozen sync that bails out still modifies the
   tree. Since `--frozen` is documented as "CI mode", it could imply `--unattended`'s no-working-tree-writes. At
   minimum the writes should move after the check.

PR #258, opened 2026-08-24
[feature] scan box import statements

rmvegasm:
# PR: `uvr scan` detects `box::use()` imports

## Summary

Teach `uvr scan` to discover package dependencies declared via the
[`box`](https://klmr.me/box) package's `box::use(...)` syntax, alongside the
existing `library()`, `require()`, `pkg::fn`, and roxygen2 detectors.

Package declarations are detected (`box::use(dplyr[filter])`,
`box::use(gg = ggplot2)`), while local module imports (`box::use(./mod)`,
`box::use(prefix/mod)`, `box::use(../mod/utils[f])`) are ignored.

## Changes

- `crates/uvr/src/commands/scan.rs`
  - Add `box_use` and `box_use_spec` regexes to `PackageDetector`.
  - Add a small `strip_r_comments` helper so `#` comments between/inside
    declarations don't break parsing.
  - Add unit tests covering single-line and multi-line `box::use` calls,
    aliases, attach lists, wildcard `...`, local module paths, and `)`
    inside comments.
- `CHANGELOG.md`
  - Add an Unreleased bullet.

## Testing

- `cargo test -p uvr commands::scan` — 8 tests, all pass
- `cargo test -p uvr --bin uvr` — 180 unit tests, all pass

## Merge note: CHANGELOG conflict

This branch is based on `main`. The other open PR (`fix/206-176`, `--no-ide`
/ `--bare`) also adds a bullet to the `## Unreleased` section of
`CHANGELOG.md`, in the same spot. Whichever PR merges second will need a
trivial conflict resolution there (keep both bullets). There is no code
overlap: the other PR does not touch `crates/uvr/src/commands/scan.rs`.


gdevenyi (commented):
Reviewed against `main`. Checked out the branch, ran the suite, clippy and rustfmt, then probed the regexes with extra cases.

**What checks out**

- `cargo fmt --all -- --check` clean, `cargo clippy --workspace --all-targets -- -D warnings` clean, 180 + 63 + 515 tests pass.
- The package-vs-module rule matches box's documented semantics. From the box vignette: "non-local module names *must* be fully qualified, nested modules: `box::use(foo/bar)` works, `box::use(bar)` does not (instead, it is assumed that `bar` refers to a *package*)". So bare name = package, anything containing `/` = module, and dropping every capture with a `/` is exactly right. There is no package-submodule case to lose.
- Multi-line attach lists work, including ones broken across lines: `box::use(\n dplyr[\n filter,\n mutate\n ],\n tidyr,\n)` yields `{box, dplyr, tidyr}`.
- `box` itself is picked up by the existing `namespace_op` regex, which is correct — a file using `box::use()` does depend on box.

**One false positive: a commented-out multi-line `box::use`**

`box_use` runs over the raw file, so a `box::use(` that only exists inside a comment still matches, and its comment-aware alternation then walks past the commented closing paren to the next real one, sweeping live code into the argument list. Minimal repro:

```r
# box::use(dplyr,
#   tidyr)
result <- transform(df)
```

gives `{box, dplyr, result}`. `result` is a local variable. It surfaces in `uvr scan` output as a missing package, and commenting out an import while debugging is common enough to hit.

**Suggested fix, which also simplifies the code**

Strip comments once before the box pass instead of after the capture. That removes the reason the outer regex needs `(?:[^)#]|#[^\n]*)*` at all:

```rust
let box_use = Regex::new(r#"\bbox\s*::\s*use\s*\(([^)]*)\)"#)
    .expect("box::use regex compiles");
```

```rust
let decommented = strip_r_comments(content);
for cap in self.box_use.captures_iter(&decommented) {
    if let Some(args) = cap.get(1) {
        for spec in self.box_use_spec.captures_iter(args.as_str()) {
```

I applied exactly this on your branch. The repro drops to `{box}`, and all four of your `box_use` tests still pass unchanged — including `extract_box_use_ignores_comment_parens`, which is what the comment-aware alternation was written for. It keeps working because the paren inside the comment is gone by the time the regex runs.

Order matters: strip only for the box pass, not before `roxygen_import`, which needs `#'` lines intact.

Worth adding the commented-out case as a fifth test either way, since it is the one that regresses if the two passes are ever reordered.

Everything else in the PR looks right to me, and the `strip_r_comments` helper carries its weight once it moves up front.

rmvegasm:
Applied the strip-comments approach: `box::use` now scans
`strip_r_comments(content)` with the simplified `\bbox\s*::\s*use\s*\(([^)]*)\)`.
Added regression tests for the commented-out block and for indented
`box::use` inside a function body. `cargo test --workspace`, clippy, and fmt
are all clean.

crates/uvr/src/commands/scan.rs
```diff
@@ -188,10 +214,30 @@ impl PackageDetector {
...
+            Regex::new(r#"\bbox\s*::\s*use\s*\(([^)]*)\)"#).expect("box::use regex compiles");
+
+        // One import declaration inside a `box::use(...)` list. Captures
+        // the raw package/module name (`pkg`, `alias = pkg`, `prefix/mod`,
+        // `./mod`, `../mod`). The trailing `\[[^\]]*\]` consumes an attach
+        // list so names inside it (`dplyr[filter, mutate]`) are never seen
+        // as packages. Callers drop captures containing `/` (module paths)
+        // and the `.`/`..` placeholders.
+        let box_use_spec = Regex::new(
+            r#"(?:^|,)\s*(?:[A-Za-z][A-Za-z0-9._]*\s*=\s*)?((?:[A-Za-z][A-Za-z0-9._]*|\.{1,2})(?:/[A-Za-z][A-Za-z0-9._]*)*)\s*(?:\[[^\]]*\])?"#,
```
gdevenyi, line 231: A relative path with more than one `..` leaks attach-list names as packages. box supports this form: `R/use.r` says "`../../` denotes the grandparent module", and `test-parse-mod-spec.r` tests `../../b[foo]` as valid.

After the first component, the path part accepts only components that start with a letter. For `../../mod[f, g]` the capture stops at `..`, so the optional `\[...\]` does not consume the attach list. The next search then starts at the `,` inside `[f, g]`:

```
box::use(../../mod[f, g])   =>  {box, g}
```

Accepting `..` as a later component fixes it:

```suggestion
            r#"(?:^|,)\s*(?:[A-Za-z][A-Za-z0-9._]*\s*=\s*)?((?:[A-Za-z][A-Za-z0-9._]*|\.{1,2})(?:/(?:[A-Za-z][A-Za-z0-9._]*|\.\.))*)\s*(?:\[[^\]]*\])?"#,
```

With this change the same input gives `{box}`, and all the existing scan tests still pass. A `../../mod[f, g]` line in `extract_box_use_imports` would guard it.

crates/uvr/src/commands/scan.rs
```diff
@@ -151,6 +152,29 @@ fn has_scannable_extension(path: &Path) -> bool {
     )
 }
 
+/// Strip R `#` comments from source, preserving newlines so line structure
+/// and the surrounding comma-separated structure survive for the box spec
+/// regex. Used by the `box::use` pass; it does not model `#` inside string
+/// literals, which is fine for the top-of-file/top-of-function statements
+/// box uses.
```
gdevenyi, line 159: Nit: this now runs over the whole file, not only over the `box::use` statements, so the reason it is safe has changed. `box::use()` arguments never contain string literals, so a `#` inside a string elsewhere only cuts its own line. That case is common: plotting code is full of colour codes like `"#1f77b4"`.

One edge case remains. A `#` string before a `box::use` on the same line hides the import:

```
x <- c("#fff"); box::use(dplyr)   =>  {box}
```

That is rare, and I would not add code for it. But the comment could say it, for example:

```suggestion
/// regex. Used by the `box::use` pass, over the whole file. It does not
/// model `#` inside string literals: `box::use()` arguments never hold
/// strings, so a `#` in a string elsewhere only truncates its own line (a
/// `box::use` later on that same line is missed).
```

gdevenyi (commented):
Re-reviewed at 05caa0e. Thanks for the quick change.

**What checks out**

- The `box::use` pass now runs on `strip_r_comments(content)`, and the simpler `[^)]*` capture is in. My repro from the last review (a commented-out multi-line `box::use` followed by live code) now gives `{box}`.
- The two new tests cover the commented-out block and an indented `box::use` in a function body.
- `cargo fmt --all -- --check` and `cargo clippy --all-targets --all-features -- -D warnings` are clean. The full suite passes: 182 + 63 + 515.
- Only `CHANGELOG.md` conflicts with current `main` (the new #261 and #271 entries). `scan.rs` merges cleanly, also with #266, which changes the same file.

**One more false positive** is in the spec regex, when a relative path has more than one `..`. The fix is one line (inline). The other inline comment is a small doc note.

Issue #252, opened 2026-08-16
Add comment about making scripts executable with a `#!/usr/bin/env -S uvr run` shebang to docs


remlapmot:
Kudos for this great package manager.

You can pull off the same trick as `uv` and make [scripts executable by adding a shebang line at the top](https://docs.astral.sh/uv/guides/scripts/#using-a-shebang-to-create-an-executable-file).

The example in your README can simply gain an additional first line, e.g.,

```r
#!/usr/bin/env -S uvr run
# /// script
# dependencies = [
#   "jsonlite",
#   "praise",
# ]
# ///

cat(praise::praise(), "\n")
```

To run, save this is in a file named _script_ and make it executable

```sh
chmod +x script
```

and you can run it with

```sh
./script
```

gdevenyi:
+1, this works today and costs one README line.

Worth flagging the portability of `env -S` alongside it, since the example will get copied:

- `-S` needs GNU coreutils 8.30 or newer (2018). CentOS/RHEL 7 ships 8.22, so `#!/usr/bin/env -S uvr run` fails there with "env: invalid option -- 'S'". Every current distro is fine.
- macOS is fine — BSD `env` has had `-S` for far longer than the GNU one.
- Windows has no shebang mechanism at all, so the line is inert there and the script still needs `uvr run script.R`. Given the README documents a Windows install path, the docs should say so rather than let people find out.

The reason `-S` is needed at all is worth one clause too: a shebang passes everything after the interpreter as a *single* argument, so `#!/usr/bin/env uvr run` looks for a program literally named `uvr run`. `-S` splits it. That is the detail people get wrong when they adapt the line.

`docs.astral.sh/uv/guides/scripts` puts this next to the inline-metadata section, which is where it belongs here too — the header and the shebang are the same feature from the user's side.

gdevenyi:
Correcting myself: most of this is already done. `fe4aa1b` ("docs: show the shebang form of standalone scripts") landed the example and the coreutils 8.30 note on 2026-08-19, so my comment above described a gap that was already closed. I read the wrong section of the README.

The one thing still missing is Windows, which has no shebang mechanism at all — the line is inert there and the script has to be run as `uvr run hooray`. Sending a one-line PR for that.

novica:
Just sharing how it has been done elsewhere 

https://a2-ai.github.io/rv-docs/reference/self_contained_scripts/

nbafrank:
[...]

PR #262, opened 2026-08-28
docs: note that the script shebang is inert on Windows (#252)

gdevenyi:
Small follow-on to `fe4aa1b`, which landed the executable-script example and the coreutils 8.30 caveat for #252.

The remaining gap is Windows. It has no shebang mechanism, so `#!/usr/bin/env -S uvr run` is ignored there and the script has to be run as `uvr run hooray`. The README documents a Windows install path, so someone will copy that example onto a Windows box and get no warning.

One sentence added to the paragraph that already covers the Linux and macOS caveats.

Docs only. No CHANGELOG entry, matching `fe4aa1b`, which was also docs-only and added none.

Leaving #252 open — the maintainer's commit message says it stays open for the reporter.


nbafrank:
Merging this! Thanks!

Issue #255, opened 2026-08-20
sync: prune deletes packages installed into the project library by other tools


nbafrank:
`uvr sync` prunes any directory in the project library that has a `DESCRIPTION`, isn't named `uvr`, and isn't in `uvr.lock` (`crates/uvr/src/commands/sync.rs:1365`). That's the right call for uvr's own leftovers, but the library is discoverable by design: `uvr activate` exports `R_LIBS_USER` pointing at it. So anything else that installs there gets silently deleted on the next sync. Two cases:

- a user running `install.packages()` inside an activated shell
- external tools that resolve the lib via `R_LIBS_USER` (carrier does exactly this)

Since only `sync` prunes and `add`/`import`/`update`/`run` don't, the deletion is intermittent. It looks like corruption rather than policy, which makes it a nasty one to debug from a bug report.

**Proposal:** spare packages carrying a foreign-install marker (a `Meta/` sentinel file, generalizing the hardcoded `name == "uvr"` companion exemption) and report them as "unmanaged, kept" instead of removing them. Maybe a `--prune-all` flag to restore today's behavior.

Came up while looking at interop with joshuamarie/carrier, but it also bites plain `install.packages()` users with no other tool involved.

gdevenyi:
Agreed this is worth fixing, but I think the sentinel is inverted as proposed.

"Spare packages carrying a foreign-install marker" requires the foreign tool to cooperate. `install.packages()` will never write a uvr marker, and neither will carrier, or renv, or a `R CMD INSTALL` by hand — which is the whole set of cases in the issue. The default for an unmarked directory would still be deletion, so nothing in the reported scenario changes.

Invert it: **mark uvr's own installs, and prune only what carries uvr's marker and is absent from `uvr.lock`.** Everything else is reported as unmanaged and kept.

That gets the failure direction right in both places that matter:

- An unmarked package is kept. Unknown provenance means leave it alone, which is the safe default for a directory uvr did not create.
- Existing project libraries have no markers, so the first sync after the change prunes nothing at all, and packages re-acquire markers as they are reinstalled. No migration step, no flag day, and none of #175's stale-artifact shape.

It also removes the special case: `name == "uvr"` stops being hardcoded at `sync.rs:1419` once the companion is marked like everything else.

[...]

Issue #247, opened 2026-08-06
perf(windows): warm cache hits do a full recursive copy — hardlink instead


nbafrank:
On Linux, warm cache hits attach via symlink; on macOS via `clonefile` (CoW). On Windows, `clone_to_library` falls through to `copy_dir_recursive` — a full physical copy of every package on every warm sync.

NTFS supports hardlinks without special privileges (`CreateHardLinkW`, `std::fs::hard_link`). Hardlinking per file gives Windows the same near-instant warm syncs as the other platforms: same-volume constraint applies (cache and project on different drives falls back to copy, as today).

R package files are effectively read-only post-install, and uv uses hardlinks for Python wheels on the same reasoning. Likely the single biggest per-platform performance gap remaining.

nbafrank:
Implemented and merged. Windows cache hits now hardlink each file into the project library instead of copying its bytes — NTFS grants hardlinks to ordinary users, unlike the symlinks that made this path a copy originally. Falls back to a copy when cache and project sit on different volumes, and cleans up any partial tree first so the fallback can't merge into it.

The tree-walking logic is tested on every platform (shape and symlink preservation everywhere, inode/nlink assertions on Unix), so a Windows-only regression can't hide — #248 tracks the Linux symlink→hardlink migration separately.

PR #263, opened 2026-08-28
Feature: attach cached packages with hardlinks on Linux (#248)

### LINKED ISSUE #248: cache(linux): hardlink files instead of symlinking the package directory
Linux attaches warm cache hits as a **directory symlink** into `.uvr/library/`. It's fast, but it has two costs worth trading away:

1. **`uvr cache clean` breaks every existing project library.** The symlinks dangle until each project's next sync reseeds the cache. Hardlinked files survive: the inode lives as long as any link does, so cleaning the cache frees only what nothing references.
2. **Tooling that resolves symlinks reports packages as living in the cache path**, not the project library — the class of confusion renv spent years fixing. `.libPaths()`, `find.package()`, and anything walking the library see through the link.

Proposal: per-file hardlinks (reflink where the filesystem supports it — btrfs/XFS — falling back to hardlink, then copy). Same dedup, same near-instant attach, none of the above. This is why uv hardlinks Python wheels.

Constraints: hardlinks require same-filesystem (cache and project library on different mounts falls back to copy); R package files are effectively read-only after install, so shared inodes are safe, but the extraction path must not ever write in place into a cached file.

Wants its own careful review — it changes what's on disk in every existing project library, and #175's ghost (stale artifacts surviving a fix) is a reminder to think about migration from existing symlinked libraries.

gdevenyi:
Fixes #248.

## What changed

Linux stops attaching a warm cache hit as a whole-directory symlink and hardlinks the files instead, which is what Windows already does.

`hardlink_dir_recursive` needed no new code. It landed for Windows in #247, and it was already compiled and unit-tested on every platform — the comment above it said so explicitly. So this is a `#[cfg]` widening on the call site plus deleting the symlink arm.

## Why

Both reasons from the issue:

1. **`uvr cache clean` no longer breaks existing project libraries.** The symlinks dangled until each project's next sync reseeded the cache. A hardlinked file outlives the cache entry, so cleaning frees only what nothing else references.
2. **`.libPaths()` and `find.package()` now report the project library**, not the cache path. That was the renv-class confusion the issue names.

Dedup and attach speed are unchanged — same shared storage, same near-instant attach.

## The tradeoff, stated plainly

A hardlink cannot cross a filesystem, so a cache and project library on different mounts now fall back to a recursive copy where a symlink used to work. That is the constraint the issue already accepted, and it is the same fallback Windows has been living with since #247.

## Migration

No flag day and no separate step. `remove_entry` already clears a stale directory symlink before attaching, so a library written by an older uvr converts one package at a time on its next sync. `prune_unused_packages` keeps its symlink branch for exactly that reason — its doc comment now says the case is historical rather than the Linux norm.

This is the #175 lesson applied: nothing has to be cleaned up by hand for the fix to take effect.

## Not in this PR

Reflink (btrfs/XFS `copy_file_range`) is deliberately left out. It changes nothing about the interface and can go in later; plain `std::fs::hard_link` already delivers both benefits above. Skipping it keeps this diff to the shape that is easy to reason about.

## On the write-in-place constraint

The issue flags that the extraction path must never write in place into a cached file. `clone_to_library` has exactly one caller (`sync.rs:662`) and nothing writes into the attached directory afterwards, so this holds today. It is worth keeping in mind for #255, where sparing foreign installs means other tools may write into the library.

## Tests

- The three Linux attach tests now assert a real directory with a **shared inode** rather than a symlink.
- One covers migrating a pre-#248 symlink in place.
- A new one removes the entire cache and then reads the library back — the regression the issue is actually about, which the old symlink attach would have failed.

`cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D warnings`, and the full suite are clean locally: 177 uvr, 63 CLI, 516 uvr-core (+1 from the new test).

## Merge note

Adds a bullet to the `## Unreleased` section of `CHANGELOG.md`, so it will conflict with the other open PRs touching that spot. Keep both bullets. No code overlap with #257 or #258.


gdevenyi:
CI note: the `Distro (rocky 10 (rhel10))` failure was environmental, not this diff. It failed in **Install host prerequisites**, before any uvr code runs:

```
Problem: cannot install the best candidate for the job
  - nothing provides libxml2(x86-64) = 2.12.5-10.el10_2.3
    needed by libxml2-devel-2.12.5-10.el10_2.3.x86_64 from appstream
```

Rocky 10's appstream mirror was briefly serving a `libxml2-devel` whose matching `libxml2` runtime had not propagated. The window was about three minutes wide and this run landed inside it — the same job passed on #262 at 15:25:27 and on #264 at 15:30:56, either side of this one at 15:28:29.

I have no rerun rights on the repository, so I refreshed the branch to trigger a new run. The tree is byte-identical to the commit you already have.

Every other job passed, macOS and Windows included.

gdevenyi:
Correcting one detail in my note above: I called it a three-minute window, but it is not time-bounded. The same failure hit #265 twenty minutes later at 15:45:25, and this branch's rerun passed at 15:46:24 — one minute after. `dnf` picks a mirror per run, so it is luck of the draw, not a window that closed.

The diagnosis is unchanged: it fails before any uvr code runs and nothing here touches that path. All 15 checks are green now.

Filed the flake separately since it will keep hitting unrelated PRs.

PR #269, opened 2026-08-31
fix(sync): pin plain git packages to their commit, not just their version

pat-s:
## The bug

`uvr sync` treats a git-sourced package as up to date when the installed **version string** matches the lockfile, without checking that the installed copy came from the **commit** the lockfile names.

For a package whose development version does not change between commits, that means it is never updated. `pat-s/cranlike` carries `Version: 1.0.3.9001` at every commit on its `s3` branch. A library holding a June build kept it through every later sync, even with the lockfile pinned to a commit two months newer.

The consequence in our CI: `bincraft` calls `cranlike::update_PACKAGES(..., built = built)`, an argument added on 2026-07-29. The stale June build has no such argument, so the R side failed with

```
Error: Failed to update PACKAGES after 3 attempts. Last error: unused argument (built = built)
```

on a runner with a persistent package cache, while a fresh runner worked. Nothing about the lockfile or the sync output suggested the installed package was two months behind what was pinned.

## Reproduction

```sh
# a library holding an older commit of the fork
R CMD INSTALL -l /tmp/lib <clone of pat-s/cranlike at a12591f>
# -> Version: 1.0.3.9001, update_PACKAGES has no `built` argument

# a project whose lockfile pins a newer commit of the same package
uvr add --no-install "forgejo::codefloe.com/rpkgs/bincraft@v5.2.1"
grep -A3 'name = "cranlike"' uvr.lock
#   source = "github"
#   url = ".../pat-s/cranlike/tarball/736d753..."
#   checksum = "git:736d753..."

uvr sync --library /tmp/lib
# installs 67 packages, never mentions cranlike
# -> still Version: 1.0.3.9001, still no `built` argument
```

## Cause

`NestedProvenance::from_locked()` returns `None` for any package without a `subdirectory`:

```rust
let Some(subdirectory) = p.subdirectory.as_deref() else {
    return Ok(None);
};
```

So no marker is written for a plain (non-nested) git package, and in `is_installed()` the provenance gate passes trivially:

```rust
(Marker::Absent, None) => true,
```

leaving only the DESCRIPTION version comparison. The lockfile already records the commit — it just is not consulted for these packages.

## Change

A root GitHub entry with a `git:<sha>` checksum now gets a marker too, recorded with an empty `subdirectory`, and `parse()` accepts that as "the repository root" rather than invalid. The existing write and compare paths in `sync.rs` then do the rest unchanged.

Deliberately scoped to GitHub, which is what `NestedProvenance` and its marker format already assume (`source=github`). GitLab and Forgejo git sources have the same gap; covering them needs the marker to carry its source, which felt like a larger change than this fix warrants. Happy to extend it if you would rather do it in one go.

## Behaviour change

Existing installs have no marker, so the first sync after upgrading reinstalls git-sourced packages once. Subsequent syncs are idempotent — verified below.

`is_installed_distinguishes_nested_identity` asserted the old contract (a root entry with no marker counts as installed) and now asserts the new one.

## Verification

With the patched binary, against the same stale library:

```
BEFORE: version 1.0.3.9001 | built arg: FALSE
  ^ cranlike 1.0.3.9001 -> 1.0.3-4.9001
  > Installing 67, upgrading 1: 65 cached - 3 from source
AFTER : version 1.0.3.9001 | built arg: TRUE

marker: source=github url=.../pat-s/cranlike/tarball/736d753... checksum=git:736d753... subdirectory=

second sync: > Installing 1 package(s)   # cranlike untouched
```

- `cargo test -p uvr-core --lib`: 519 passed
- `cargo test -p uvr --bins`: 177 passed
- `cargo fmt --check`, `cargo clippy`: clean
- Four new unit tests: a plain GitHub package is pinned; a non-GitHub source is ignored; an entry without a commit is ignored; a marker round-trips an empty subdirectory.

### DIFF
--- a/crates/uvr-core/src/installer/nested_source.rs
+++ b/crates/uvr-core/src/installer/nested_source.rs
@@ -15,13 +15,33 @@ pub struct NestedProvenance {
 impl NestedProvenance {
     pub fn from_locked(p: &LockedPackage) -> Result<Option<Self>> {
         crate::registry::github::validate_nested_lock_entry(p)?;
-        let Some(subdirectory) = p.subdirectory.as_deref() else {
-            return Ok(None);
+        let subdirectory = match p.subdirectory.as_deref() {
+            Some(subdirectory) => subdirectory.to_string(),
+            // A plain (non-nested) GitHub package still needs its commit
+            // pinned. Without a marker the only thing left to compare is the
+            // DESCRIPTION version, and a package that carries the same
+            // development version across commits - `1.0.3.9001` for every
+            // commit of a fork, say - then looks up to date forever, no matter
+            // which commit the lockfile names.
+            None => {
+                if p.source != crate::lockfile::PackageSource::GitHub {
+                    return Ok(None);
+                }
+                let has_commit = p
+                    .checksum
+                    .as_deref()
+                    .and_then(|c| c.strip_prefix("git:"))
+                    .is_some_and(crate::registry::github::is_full_commit_sha);
+                if !has_commit || p.url.as_deref().unwrap_or_default().is_empty() {
+                    return Ok(None);
+                }
+                String::new()
+            }
         };
         Ok(Some(NestedProvenance {
             url: p.url.clone().unwrap_or_default(),
             checksum: p.checksum.clone().unwrap_or_default(),
-            subdirectory: subdirectory.to_string(),
+            subdirectory,
         }))
     }
 
@@ -60,7 +80,12 @@ impl NestedProvenance {
             *slot = Some(value.to_string());
         }
         let (source, url, checksum, subdirectory) = (source?, url?, checksum?, subdirectory?);
-        if source != "github" || url.is_empty() || !crate::subdirectory::is_valid(&subdirectory) {
+        // An empty subdirectory is the repository root: a plain GitHub package
+        // rather than one nested in a monorepo.
+        if source != "github"
+            || url.is_empty()
+            || (!subdirectory.is_empty() && !crate::subdirectory::is_valid(&subdirectory))
+        {
             return None;
         }
         if !checksum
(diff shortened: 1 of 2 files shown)


crates/uvr-core/src/installer/nested_source.rs
```diff
@@ -15,13 +15,33 @@ pub struct NestedProvenance {
...
-            return Ok(None);
+        let subdirectory = match p.subdirectory.as_deref() {
+            Some(subdirectory) => subdirectory.to_string(),
+            // A plain (non-nested) GitHub package still needs its commit
+            // pinned. Without a marker the only thing left to compare is the
+            // DESCRIPTION version, and a package that carries the same
+            // development version across commits - `1.0.3.9001` for every
+            // commit of a fork, say - then looks up to date forever, no matter
+            // which commit the lockfile names.
+            None => {
```
gdevenyi, line 26: This makes a root GitHub package provenance-pinned, but `select_pkg_plan` (`sync.rs:86`) still forces a source build only when `subdirectory.is_some()`. A root GitHub package whose name and version are also in a P3M or custom-repository index still gets that registry binary. The install path then writes a marker that says the locked commit is installed.

I checked this with a throwaway test on this branch: `nested_locked("rlang", NESTED_SHA, None)` plus a custom registry that lists rlang 1.1.6 gives

```
PROBE is_binary=true url=https://rpkgs.example.com/src/contrib/rlang_1.1.6.tar.gz provenance=Some("git:0123456789abcdef0123456789abcdef01234567")
```

So a patched fork that keeps the CRAN version string gets the CRAN binary, and from then on `is_installed` reports the locked commit as present. Before this PR, the same wrong install was at least not claimed as pinned.

A smaller effect of the same mismatch: such a package is stored in the package cache under a binary key, but looked up under a source key (`binary_cache_allowed && nested.is_none()` is now `false`), so a fresh library downloads it again.

The smallest fix is to force source wherever the code now treats the package as pinned, in `select_pkg_plan` and in the `detected_per_plan` closure (`sync.rs:870`). The existing comment already gives the reason ("a same-name binary is a different package"):

```rust
if p.subdirectory.is_some() || matches!(NestedProvenance::from_locked(p), Ok(Some(_))) {
```

A test next to `select_plan_forces_source_for_a_nested_package`, with `nested_locked("rlang", NESTED_SHA, None)`, would guard it.

gdevenyi (commented):
Reviewed at a31c7ae against current `main`. I built the branch and ran `cargo fmt --all -- --check`, `cargo clippy --all-targets --all-features -- -D warnings` and `cargo test --all`: all clean (177 + 63 + 519). The branch merges into `main` with no conflicts.

**What checks out**

- The cause is right: `NestedProvenance::from_locked` returned `None` for a package without a subdirectory, so `provenance_matches` passed on `(Absent, None)` and only the DESCRIPTION version was compared.
- The fix reuses the existing marker, so the write and compare paths in `sync.rs` need no change. The empty-subdirectory marker round-trips, and the updated `is_installed_distinguishes_nested_identity` pins the new rule, including a root entry at another commit.

**One gap:** a root GitHub package can still be installed from a registry binary, and it now gets a commit marker. Details inline. The registry-binary swap itself already exists on `main`; I opened #300 for it.

**Other notes**

- **Cache key.** For a root GitHub package, the package-cache key now uses `cache_identity()` (`github|<url>|<checksum>|`) instead of the bare checksum (`sync.rs:654` and `:1332`), and `provenance_matches` also checks the marker in the cached entry. Existing cache entries for these packages therefore no longer match. The one-time reinstall that the PR body describes is a full download and source build, not a cache clone, and the old entries stay in the cache until `uvr cache clean`. That is acceptable, but worth one line in the PR body.
- **CHANGELOG.** This is a user-visible fix with a one-time reinstall, so a bullet in `## Unreleased` would help.
- **GitLab and Forgejo.** Agreed that the marker needs a source field for them. Open PR #298 adds a `GitHost` type for all three hosts, which could carry it later.
- **Related work.** #277 makes `uvr sync --frozen` compare the locked commit of git sources with a fresh resolve. This PR compares the installed copy with the locked commit. The two are complementary and do not conflict.

PR #273, opened 2026-09-16
feat(sysreqs): recognize RPM providers and configurable alternatives

pat-s:
System-requirements catalogs can name a compatibility package even when an image already supplies a verified alternative.
On RHEL 9, for example, requesting `gdal3.4-devel` conflicts with the installed canonical `gdal-devel`, although the SDK is usable.

Recognize declared RPM providers with `rpm -q --whatprovides` and add operator-supplied, distro-scoped installed-provider overrides through `UVR_SYSREQS_OVERRIDES`.
An override accepts only an already-installed alternative, optionally guarded by a `pkg-config` module and minimum SDK version.
The resolver retains the original requirement when a provider is absent or a guard fails, rejects malformed policy files, and prints each applied override.
Both the Posit API and local-rule paths use the same policy before collecting installation/setup commands.
There is no GDAL-specific production-code exception and no automatic package replacement or erasure.

For the motivating case, an image can declare:

```toml
[[overrides]]
distro = "redhat-9"
package = "gdal3.4-devel"
installed = "gdal-devel"
pkg_config = "gdal"
minimum_version = "3.4"
```

The separate `gdal3.4` runtime requirement remains intact: the binary `sf` and `terra` packages tested here require `libgdal.so.30`.
Version guards use `pkg-config` rather than interpreting arbitrary RPM versions as semantic versions.
The dependency-check API now returns `Result` so explicit configuration errors stop sync.

Changed files:

- `crates/uvr-core/src/lib.rs`: register the overrides module.
- `crates/uvr-core/src/sysreqs.rs`: recognize RPM providers, apply policy to both resolution paths, and test setup-command filtering and deduplication.
- `crates/uvr-core/src/sysreqs_overrides.rs`: load, validate, and evaluate scoped policy with optional SDK checks.
- `crates/uvr-core/tests/sysreqs_rpm.rs`: cover virtual providers, failed guards, missing tools/providers, wrong distro, missing or malformed policy, and runtime preservation in isolated subprocesses.
- `crates/uvr/src/commands/sync.rs`: propagate configuration errors and report applied overrides.

Validation:

- `cargo test`: 768 passed, 15 ignored.
- `cargo fmt --all --check` and `cargo clippy --all-targets --all-features -- -D warnings`: passed.
- All current upstream PR checks passed, including manylinux and musllinux ARM64, Windows, macOS, and the distro checks.
- Build the pinned commit into a static musl binary using an Alpine Rust container.
- Apply that binary and the example policy to the published RHEL 9 amd64 build image: all 12 R dependencies of the smoke project install, `sf`, `s2`, `units`, and `terra` load, and basic spatial operations pass under R 4.6.0.
- Verify GDAL 3.10.3 development metadata remains selected while GDAL 3.4 runtime libraries satisfy the R binaries.
- Start the same static binary successfully on UBI 8 and UBI 10.

The RHEL container smoke test was on amd64; a full clean multiarchitecture build of the downstream environment images remains pending.


crates/uvr-core/src/sysreqs.rs
```diff
@@ -571,33 +571,62 @@ async fn fetch_sysreqs_index(client: &reqwest::Client, distro: &str) -> SysReqIn
...
+}
+
+fn query_installed(cmd: &str, args: &[&str], package: &str) -> bool {
+    std::process::Command::new(cmd)
+        .args(args)
+        .arg(package)
+        .stdout(std::process::Stdio::null())
+        .stderr(std::process::Stdio::null())
+        .status()
+        .is_ok_and(|status| status.success())
```
gdevenyi, line 593: This changes what a probe that cannot start means. The old code in `filter_missing` had `Err(_) => false, // command failed to run — don't report as missing`: if `which` found the package manager but it could not be started, uvr reported nothing as missing. Now `query_installed` returns `false` for a spawn error, so `filter_missing` reports **every** requirement as missing, and `--install-system-deps` would then try to install all of them.

For the override check, `false` on a spawn error is the right, careful answer. For `filter_missing` it is the opposite. One way to keep both:

```rust
fn query_installed(cmd: &str, args: &[&str], package: &str) -> Option<bool> {
    std::process::Command::new(cmd)
        .args(args)
        .arg(package)
        .stdout(std::process::Stdio::null())
        .stderr(std::process::Stdio::null())
        .status()
        .ok()
        .map(|status| status.success())
}
```

Then `filter_missing` keeps a requirement only when the answer is `Some(false)`, and `has_installed_package` accepts only `Some(true)`.

crates/uvr-core/src/sysreqs_overrides.rs
```diff
@@ -0,0 +1,176 @@
...
+        let mut seen = HashSet::new();
+        for rule in &policy.overrides {
+            for (field, value) in [
+                ("distro", rule.distro.as_str()),
+                ("package", rule.package.as_str()),
+                ("installed", rule.installed.as_str()),
+            ] {
+                validate_token(field, value)?;
+            }
+            if !rule.distro.contains('-') {
```
gdevenyi, line 54: `distro` is compared with the value from `detect_linux_distro()`, which uses uvr's normalized names, not the os-release values (`normalize_distro` in `sysreqs.rs`):

- RHEL and Oracle Linux 9.x are `redhat-9`.
- Rocky Linux and AlmaLinux 9 are `rockylinux-9`.
- Only the RHEL family is cut to its major version, so openSUSE stays `opensuse-15.6`.

An operator who writes the os-release spelling (`rhel-9`, `rocky-9`, `almalinux-9`) or the full RHEL version (`redhat-9.4`) gets a valid policy that never matches, and uvr prints nothing.

Two options:

- Normalize the configured value when the policy is parsed: split with `rsplit_once('-')` (an os-release ID such as `opensuse-leap` contains a `-`, a version does not) and pass both halves through `normalize_distro`. Then both spellings work.
- Keep the strict form, but document the accepted names and warn when no rule's `distro` equals the host's value.

Either way, the accepted spelling needs to be written down (see the other comment).

crates/uvr-core/src/sysreqs_overrides.rs
```diff
@@ -0,0 +1,176 @@
...
+    distro: String,
+    package: String,
+    installed: String,
+    pkg_config: Option<String>,
+    minimum_version: Option<String>,
+}
+
+impl Overrides {
+    pub(crate) fn load() -> Result<Self> {
+        let Some(path) = std::env::var_os("UVR_SYSREQS_OVERRIDES").filter(|p| !p.is_empty()) else {
```
gdevenyi, line 29: Two things about this variable:

- **No documentation.** It is a new user-facing setting, but after a8aa98d ("remove README changes") nothing describes it: not the README, not `CHANGELOG.md`, and not `env_vars.rs`. An operator needs the file format and the `distro` spelling to use it.
- **Where it is read.** Most `UVR_*` variables are read in `crates/uvr-core/src/env_vars.rs`, each with a doc comment and a line in `test_env_vars`, through `read_env_var`. That function also treats a whitespace-only value as unset; `var_os(...).filter(|p| !p.is_empty())` does not, so `UVR_SYSREQS_OVERRIDES=" "` becomes a failed read of a file named ` `. A `sysreqs_overrides() -> Option<PathBuf>` helper there would follow the existing pattern.

crates/uvr-core/src/sysreqs.rs
```diff
@@ -684,7 +715,8 @@ pub async fn check_system_deps(
     client: &reqwest::Client,
     packages: &[PackageSysReqQuery],
     distro: &str,
-) -> SysReqsCheck {
+) -> anyhow::Result<SysReqsCheck> {
```
gdevenyi, line 718: Nit: this is the first `anyhow` type in `uvr-core`'s source and public API. The rest of the crate returns `crate::error::Result` / `UvrError` and leaves `anyhow` to the CLI crate. `UvrError::Other(format!(...))` (or a new variant) would keep `check_system_deps` consistent with the functions around it, and `sysreqs_overrides.rs` could use the same type.

crates/uvr/src/commands/sync.rs
```diff
@@ -998,7 +998,10 @@ async fn install_from_lockfile_with_r(
                     .collect();
 
                 if !queries.is_empty() {
-                    let check = sysreqs::check_system_deps(&client, &queries, &distro).await;
+                    let check = sysreqs::check_system_deps(&client, &queries, &distro).await?;
+                    for description in &check.overrides_applied {
+                        eprintln!("System dependency override: {description}");
```
gdevenyi, line 1003: Nit: the rest of this block reports through `ui::` (`ui::warn_block`, `ui::hint`), which keeps the symbols and colours consistent. `ui::bullet_dim(format!("System dependency override: {description}"))` or `ui::info(...)` would match.

gdevenyi (commented):
Reviewed at a8aa98d against current `main`. I built the branch and ran `cargo fmt --all -- --check`, `cargo clippy --all-targets --all-features -- -D warnings` and `cargo test --all`: all clean (177 + 63 + 526, plus the 2 new `sysreqs_rpm` tests). The branch merges into `main` with no conflicts.

**What checks out**

- The policy is careful in the right direction. An override drops a requirement only when the named alternative is installed and, if configured, `pkg-config --atleast-version` agrees. A missing `pkg-config`, a missing provider or a failed guard keeps the original requirement, and nothing is erased or replaced.
- Policy values are validated before use: they cannot start with `-` or contain `/` or shell syntax, and the tools run through `Command` with no shell. A duplicate rule or an unknown key is an error, and a bad policy file stops the sync instead of being ignored.
- Both resolution paths (Posit API and local rules) go through `filter_with_overrides`, and setup commands are dropped when nothing is left missing. A test covers this.
- `sysreqs_rpm.rs` sets `PATH` in a child process, which is a good way to fake `rpm` and `pkg-config` without affecting other tests.

**Main points** (inline)

1. `query_installed` changes what a probe that cannot start means, on every host, not only with a policy file.
2. The `distro` key must be uvr's normalized name (`redhat-9`, `rockylinux-9`), and a wrong spelling never matches, with no message.
3. `UVR_SYSREQS_OVERRIDES` has no documentation now (the last commit removed the README text), and it is read outside `env_vars.rs`.

**Smaller items**

- **`rpm -q --whatprovides`** now applies on every RPM host, with or without a policy file. I think that is an improvement, because a virtual provider is a real provider. But it is a behaviour change, so it deserves one line in the PR body and a CHANGELOG bullet.
- **A test that can pass without checking anything.** `api_overrides_are_deduplicated_and_drop_unneeded_setup_commands` returns early when `bash` is not visible through dpkg, rpm or apk. On my machine (Arch-based, none of the three) it passes without asserting anything. The child-process fake `rpm` from `sysreqs_rpm.rs` would make it run everywhere.
- **Merge note.** Open PR #265 (`uvr sysdeps`) also calls `check_system_deps`, and `git merge-tree` shows a conflict in `sysreqs.rs`. Whichever PR merges second needs the new `Result`; I can update #265 if this one merges first.

pat-s:
The failing CI check is unrelated to this PR.

